Issue #5, opened 2019-01-02
Intermittent Yarn issues


rhlsthrm:
We are seeing a variety of yarn related issues mostly from running `yarn start`. Some of the errors we have seen include:

```
warning machinomy > @types/commander@2.12.2: This is a stub types definition for commander (https://github.com/tj/commander.js). commander provides its own type definitions, so you don't need @types/commander installed!
warning machinomy > @machinomy/contracts > @machinomy/hdwallet-provider > web3-provider-engine > ethereumjs-block > merkle-patricia-tree > level-ws > xtend > object-keys@0.4.0:
[2/4] Fetching packages...
[3/4] Linking dependencies...
error An unexpected error occurred: "ENOENT: no such file or directory, lstat '/root/.cache/yarn/v4/npm-@machinomy-hdwallet-provider-1.3.1-a7ecf4b70dfb1f59d550856b8e1884ac0ff6591f/node_modules/@machinomy/hdwallet-provider/
```

```
=============
[Makefile] => Start building hub-node-modules
docker run --name=connext_buidler --tty --rm --user 501:20 --volume=/Users/rahul/Desktop/connext/connext/indra/modules/hub:/root connext_builder:dev 501:20 "yarn install --network-timeout 1000000"
Running command as user 501:20
yarn install v1.12.3
info No lockfile found.
[1/4] Resolving packages...
warning connext > web3 > web3-bzz > swarm-js > fs-promise@2.0.3: Use mz or fs-extra^3.0 with Promise Support
warning connext > web3 > web3-bzz > swarm-js > tar.gz@1.0.7: ⚠️  WARNING ⚠️ tar.gz module has been deprecated and your application is vulnerable. Please use tar module instead: https://npmjs.com/tar
warning machinomy > @types/commander@2.12.2: This is a stub types definition for commander (https://github.com/tj/commander.js). commander provides its own type definitions, so you don't need @types/commander installed!
error Received malformed response from registry for "mocha". The registry may be down.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Same user, skipping permission fix
make: *** [hub-node-modules] Error 1
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

```
[2/4] Fetching packages...
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
[...]

PR #30, opened 2019-01-10
Extract client

bohendo:
## Types of changes

## Checklist:


bohendo:
TL;DR (Too long, didn't review):

`rm -rf modules/hub/src/vendor` and then change all the connext imports so they pull connext code from node_modules. This also required an explicit import of a bignumber library and an update to lots of bignumber operations (eg add -> plus). A lot of the changes in the hub code are just tweaking & reorganizing imports.

bohendo:
Also, all the yarn stuff has been replaced with npm and our `yarn install` errors have magically disappeared.

The `npm ln` (formerly `yarn link`) setup is now done in the Makefile rather than scattered among random entrypoint scripts. Had to massage some stuff in the top-level `ops` to get that working.

ArjunBhuptani:
Fixes #5

modules/hub/ops/prod.dockerfile
```diff
@@ -1,7 +1,9 @@
 FROM connext_builder:dev
 
-COPY node_modules node_modules
-COPY ops ops
-COPY dist dist
+COPY modules/client /client
```
bohendo: This is temporary, once we have a better way to keep the connext module on npm up-to-date, we won't need to manually copy the client module into the hub's prod-mode image

modules/hub/package.json
```diff
@@ -64,18 +65,15 @@
...
     "node-fetch": "^2.0.0",
-    "node-sass-chokidar": "^0.0.3",
     "nodemon": "^1.18.9",
     "path-to-regexp": "^2.1.0",
     "pg": "^7.4.1",
     "pg-types": "^1.13.0",
     "sql-template-strings": "^2.2.2",
     "uuid": "^3.2.1",
-    "web3": "1.0.0-beta.34",
-    "web3-old": "npm:web3@^0.20.1",
```
bohendo: npm doesn't understand this syntax, not sure what this was supposed to do but it doesn't seem to be necessary.

modules/hub/src/ChainsawService.ts
```diff
@@ -1,19 +1,18 @@
+import { BigNumber } from 'bignumber.js'
```
bohendo: I occasionally put imports into alphabetical order which makes the diff look messier than it actually is

modules/hub/src/domain/WithPayment.ts
```diff
@@ -1,5 +0,0 @@
-import Payment from 'machinomy/dist/lib/payment'
-
-export interface WithPayment {
-  payment: Payment
-}
```
bohendo: I removed the machinomy import while trying to find a broken package. This module isn't used anywhere so I never put it back after I realized the bug was elsewhere..

ArjunBhuptani: Nice, I had made a note to remove it somewhere anyway :smile:

PR #43, opened 2019-01-15
Extract client

rhlsthrm:
Updates code in extract client branch to latest Spankchain code. Fixes for dependencies and BigNumber issues. I think we will have to re-add the other changes with the imports. We are planning to do this on the spank side as well so we should coordinate the change with them.


Makefile
```diff
@@ -105,40 +107,30 @@ proxy: $(shell find $(proxy) $(find_options))
 wallet-prod: wallet-node-modules $(shell find $(wallet)/src $(find_options))
 	$(log_start)
 	$(docker_run_in_wallet) "rm -f .env && cp ops/prod.env .env"
```
bohendo: Switching to `rm -rf` should make switching from other branches easier w/out causing any other problems right? But I see this change was reverted in commit 4f713f9, was that bc it caused some problems?

modules/hub/src/CRAuthManager.ts
```diff
@@ -43,7 +43,7 @@ export class MemoryCRAuthManager implements CRAuthManager {
     const hash = this.sha3(`${MemoryCRAuthManager.HASH_PREAMBLE} ${this.sha3(nonce)} ${this.sha3(origin)}`)
     const sigAddr = this.extractAddress(hash, signature)
 
-    if (!sigAddr || sigAddr.toLowerCase() !== address.toLowerCase()) {
+    if (!sigAddr || sigAddr !== address) {
```
bohendo: Do these toLowerCase casts happen elsewhere then?

modules/hub/src/vendor/connext/hasPendingOps.ts
```diff
@@ -0,0 +1,12 @@
...
+export function hasPendingOps(state: ChannelState) {
+  for (let field in state) {
+    if (!field.startsWith('pending'))
+      continue
+    if ((state as any)[field] != '0')
+      return true
+  }
+  return false
+}
+
```
bohendo: Are changes in `modules/hub/src/vendor/connext` still necessary? If so, is the client even really extracted?

Looks like we're importing from vendor rather than node_modules, I'm curious about why this is.

ArjunBhuptani:
I dont think this PR actually extracts the client either. I think there was a lot lost in translation (i.e. updating with spank code) here.

@bohendo can we import by cloning the client repo instead of using npm? Would be easier to maintain for now and it seems like it's a better workflow for making client changes in lockstep with indra.

@ConnextProject/dev I'm a little frustrated about this. Seems like people's time is being wasted and their work overwritten. Can we please consolidate asap? Either we fix this problem now or we stop maintaining indra until we do, because doing both is not working

Makefile
```diff
@@ -105,40 +107,30 @@ proxy: $(shell find $(proxy) $(find_options))
...
 wallet: wallet-node-modules $(shell find $(wallet)/src $(find_options))
 	$(log_start)
 	docker build --file $(wallet)/ops/dev.dockerfile --tag $(project)_wallet:dev $(wallet)
 	$(log_finish) && touch build/wallet
 
-wallet-node-modules: $(project)_builder $(wallet)/package.json client
+wallet-node-modules: builder client $(wallet)/package.json
 	$(log_start)
-	$(docker_run_in_wallet) "yarn install --network-timeout 1000000"
+	$(docker_run_in_wallet) "rm -f node_modules/connext"
```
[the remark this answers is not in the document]
rhlsthrm: I switched it back because cleaning worked and I didn't want to make any unnecessary downstream issues. Please push a fix if you think it should be a different way.

modules/hub/src/CRAuthManager.ts
rhlsthrm: Yeah, we didn't have to do this inside the Spank hub, so we need to make sure we are doing it on the wallet side.

modules/hub/src/vendor/connext/hasPendingOps.ts
rhlsthrm: Yep this PR overwrote your changes. We need to make this change from the Spank side so that we can keep the code consistent from there. We have a PR in there which we will bring into here when it gets merged.

rhlsthrm:
@ArjunBhuptani totally agree. It seems like it's been a miscommunication that we should never be making changes to this repo until we can get Spankchain using this code as the source of truth (in progress). The problem is when we make hub changes in Indra they will have to be overwritten (or merged somehow) by the Spankchain camsite hub code. It's not as straightforward in that case as the nested git repo idea, because the hub is not in its own repo, so there will be a lot of extraneous files that need to be moved around and restructured. The best plan for now I think is to not make changes in the hub source here and use their repo as the source of truth while we transition them to using Indra.

rhlsthrm:
I think for now the best thing is not to make changes in the hub module of Indra. We should get the source from Spankchain hub until we get them on Indra.

Issue #66, opened 2019-01-25
Local client dev in wallet


rhlsthrm:
Wallet is currently using the NPM version of the Connext package. This is probably okay for now, but we may want to have yalc or some way to dev with a local client.

bohendo:
This is wired up when deploying in dev-mode w docker. We can reopen if this is still an issue in other contexts

PR #74, opened 2019-02-02
Small Devops Tweaks

bohendo:
## Description

Fixes a few small errors in the Makefile. A couple pieces of the deployment script have been polished.

Fixes issue #66

## Motivation and Context

Seems like docker isn't the ideal dev env for everyone. Luckily, a docker deployment can exist alongside a local one so even if I'm the only one benefiting from these fixes to the docker deployment, these changes still shouldn't get in anyone else's way. 

## How Has This Been Tested?

`npm start` runs without errors.

The UI loads and I can deposit without errors.. Except my deposit never seems to be registered.. I think if there are errors in the deposit function they were not introduced by these code changes though. Hard to tell, I'll dig deeper.

## Types of changes

## Checklist:


bohendo:
OH! Very Important Thing I just noticed. Assuming Indra is going to be the source of truth for the client, be aware that this PR also checks in all the client code.

The connext-client source was being ignored by the Indra repo since it was being treated as a submodule (effectively the same as having `modules/client` in the .gitignore). If we want our changes to client to persist in Indra's repo history, then checking this code in is very important.

Makefile
```diff
@@ -47,16 +47,20 @@ dev: database ethprovider hub wallet proxy
...
 	docker container prune -f
 
 reset: stop
-	docker volume rm connext_chain_dev || true
-	rm -f $(contracts)/build/state-hash
-	docker volume rm connext_database_dev || true
+	docker volume rm connext_database_dev connext_chain_dev || true
 	docker volume rm `docker volume ls -q | grep "[0-9a-f]\{64\}" | tr '\n' ' '` 2> /dev/null || true
 
+reset-client:
```
bohendo: We don't need to do this every time we build. For now, keep it available as a separate rule.

ops/deploy.dev.sh
```diff
@@ -6,7 +6,7 @@ set -e
 
 # set any of these to "yes" to turn on watchers
 watch_ethprovider="no"
-watch_hub="yes"
+watch_hub="no"
```
bohendo: These are expensive (in terms of cpu usage & time to start-up) so let's keep them off by default & let devs selectively turn them on as needed.

ops/unlock-db.sh
```diff
@@ -1,9 +1,15 @@
...
-service=connext_database
-container=`for f in $(docker service ps -q $service)
-do docker inspect --format '{{.Status.ContainerStatus.ContainerID}}' $f
-done | head -n1`
+project=connext
+service=${project}_database
 
-docker exec $container rm /var/lib/postgresql/data/postmaster.pid
+service_id="`docker service ps -q $service | head -n 1`"
+container_id="`docker inspect --format '{{.Status.ContainerStatus.ContainerID}}' $service_id`"
```
bohendo: Same functionality as before but now it's more readable

PR #76, opened 2019-02-05
Fixed Auth Hashing

base0010:
Buffers Buffer Objects and spaces getting passed into hashing function
## Types of changes

## Checklist:

### DIFF
--- a/modules/hub/src/CRAuthManager.ts
+++ b/modules/hub/src/CRAuthManager.ts
@@ -40,7 +40,7 @@ export class MemoryCRAuthManager implements CRAuthManager {
       return null
     }
 
-    const hash = this.sha3(`${MemoryCRAuthManager.HASH_PREAMBLE} ${this.sha3(nonce)} ${this.sha3(origin)}`)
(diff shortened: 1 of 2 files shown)


base0010:
@rhlsthrm

rhlsthrm:
Great stuff @base0010. I think the other change will have to be made inside modules/client, which does the front end auth, right?

base0010:
Yeah that commit includes the change to AuthenticationClient (Covers MetaMask)  autosigner dosen't append the ETH message prefix afaik.

ArjunBhuptani:
@rhlsthrm Can we include this fix in the current sprint? It might mean tweaking the signer service again but we should do it BEFORE users generate sigs that hold funds.

rhlsthrm:
Hey @base0010, this breaks our current clients with the spaces removed. I believe your issue should be fixed by this commit: https://github.com/ConnextProject/indra/commit/ff10cb934c6afecb80036ff5d70e4c159df8e57e, but you'll have to change your client to add the spaces back in.

base0010:
@rhlsthrm the spaces shouldn't be a problem to add back, def want to be explicit about that in the documentation going forward. I'll double check that this hashes the correct data with my client.

PR #86, opened 2019-02-25
Config endpoint

bohendo:
## Description

Hub: Make it's eth network id available via the config endpoint
Client: Updates to getConnextClient to correctly fetch config from hub and merge them into provided options.
Wallet: remove all the addresses from env vars, it gets these from the hub now.

## Motivation and Context

Now our client can be used for more than one set of hub/contract addresses. Also opens up the possibility of switching hubs dynamically at run-time via user input, etc.

## How Has This Been Tested?

CI approved (eventually, hopefully)

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] I have updated the documentation accordingly.


modules/client/src/Connext.ts
```diff
@@ -98,17 +99,23 @@ export interface IHubAPIClient {
...
   private user: Address
   private networking: Networking
-  private tokenName?: string
+  // private tokenName?: string
 
   constructor(user: Address, networking: Networking, tokenName?: string) {
     this.user = user
     this.networking = networking
-    this.tokenName = tokenName
+    // this.tokenName = tokenName
```
LayneHaber: should delete the reference here if were not using it, minor tweak

modules/client/src/Connext.ts
```diff
@@ -766,12 +762,36 @@ export interface ConnextClientOptions {
   contract?: IChannelManager
 }
 
+function hubConfigToClientOpts(config: HubConfig) {
```
LayneHaber: should be in types, should probably move the client options into types as well

modules/client/src/controllers/BuyController.test.ts
```diff
@@ -28,7 +28,7 @@ describe('BuyController: unit tests', () => {
   //         amount: { amountToken: '1', amountWei: '0' },
   //         type: 'PT_CHANNEL',
   //         meta: {},
-  //         recipient: '$$HUB$$',
+  //         recipient: process.env.HUB_ADDRESS,
```
LayneHaber: @rhlsthrm so i know this was a hack to inject the hub address, changing this wouldnt cause any issues hub side right?

modules/client/src/lib/constants.ts
```diff
@@ -36,6 +36,8 @@ export const BOOTY = {
   type: CurrencyType.BEI,
 }
 
+export const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000'
```
LayneHaber: the utils file in the client exports an emptyAddress and emptyRootHash (uses it for some generation fns). should probably keep all the constants in here though, since it looks like the hub is importing the constants file from the lib of the client as well

modules/client/src/Connext.ts
rhlsthrm: Yea, I've seen this not being used for a while now.

PR #88, opened 2019-02-27
Linked payments

LayneHaber:
## Description
Can redeem payments from link/qr code

## Types of changes


modules/client/src/controllers/RedeemController.test.ts
```diff
@@ -0,0 +1,59 @@
...
+          pendingDepositTokenUser: '1',
+          pendingDepositWeiHub: '0',
+          pendingDepositWeiUser: '0',
+        },
+      },
+    })
+
+  })
+
+  it('should fail if invalid secret is provided', async () => {
```
rhlsthrm: What if it is a hex string but not the right hex string? Do we need a different test for that?

modules/client/src/controllers/RedeemController.ts
```diff
@@ -0,0 +1,20 @@
...
+    if (!Web3.utils.isHex(secret)) {
+      throw new Error(`The secret provided is not a hex string. Was it generated using the 'generateSecret' method of connext? Secret: ${secret}`)
+    }
+
+    try {
+      const res = await this.hub.redeem(secret)
+      this.connext.syncController.handleHubSync(res.sync)
+      return { purchaseId: res.purchaseId }
+    } catch (e) {
+      throw new Error(`Error redeeming payment with secret: ${secret}` + e.message)
```
rhlsthrm: Seems like this is the error that should be triggered in one of the above tests.

modules/database/migrations/sqls/20190222054946-add-payment-secret-up.sql
```diff
@@ -0,0 +1,63 @@
...
+  p.id,
+  up.created_on,
+  up.contract,
+  p.purchase_id,
+  up. "user" AS sender,
+  p.recipient,
+  p.amount_wei,
+  p.amount_token,
+  p.meta,
+  CASE WHEN p.secret IS NOT NULL
```
rhlsthrm: Nice!

modules/hub/src/PaymentsService.ts
```diff
@@ -131,9 +151,13 @@ export default class PaymentsService {
         await this.doInstantCustodialPayment(p.payment, p.paymentId)
       } finally {
         // Check to see if collateral is needed, even if the tip failed
-        const [res, err] = await maybe(this.channelsService.doCollateralizeIfNecessary(p.payment.recipient))
-        if (err) {
-          LOG.error(`Error recollateralizing ${p.payment.recipient}: ${'' + err}\n${err.stack}`)
+        // if the payment isnt going to an empty addr (as is the) case
+        // for PT_LINK
+        if (p.payment.recipient !== emptyAddress) {
+          const [res, err] = await maybe(this.channelsService.doCollateralizeIfNecessary(p.payment.recipient))
```
rhlsthrm: Fancy type stuff 🐙

modules/hub/src/PaymentsService.ts
```diff
@@ -144,6 +168,75 @@ export default class PaymentsService {
...
+    try {
+      if (purchasePayment) {
+        await this.doInstantCustodialPayment(purchasePayment, payment.id)
+        // mark the payment as redeemed by updating the recipient field
+        const redeemedPaymentRow = await this.paymentMetaDao.redeemLinkedPayment(user, secret)
+        purchaseId = redeemedPaymentRow.purchaseId
+      }
+    } finally {
+      // always check for collateralization
+      const [res, err] = await maybe(this.channelsService.doCollateralizeIfNecessary(payment.recipient))
```
rhlsthrm: Why don't we just do this before instead of in the `finally`?

modules/hub/src/PaymentsService.ts
```diff
@@ -144,6 +168,75 @@ export default class PaymentsService {
...
+      throw new Error('Channel is not open: ' + user)
+    }
+
+    const payment = await this.paymentMetaDao.getLinkedPayment(secret)
+    // if hub can afford the payment, sign and forward payment
+    // otherwise, collateralize the channel
+    const prev = convertChannelState('bn', channel.state)
+    const amt = convertPayment('bn', payment.amount)
+    let purchasePayment = null
+    if (!this.validator.cantAffordFromBalance(prev, amt, "hub")) {
```
rhlsthrm: This is a little confusing with the double negative. Why don't we do something like:
```
// doCollateralizeIfNecessary here
if (this.validator.cantAffordFromBalance(prev, amt, "hub")) {
  // return null here, or return an error
}

// rest of function assuming happy case
```

rhlsthrm: Something Slipper told me one time: It's better and more readable to not nest the happy case stuff in `if` statements, use the `if`s to check invariants and return/error out of the function early, so the stuff outside of the if's is all the normal/happy case stuff.

modules/hub/src/PaymentsService.ts
rhlsthrm: Also you need to await this.

modules/hub/src/vendor/connext/types.ts
```diff
@@ -7,9 +7,9 @@ export type Address = string
 
 // alias functions
 // @ts-ignore
-export const isBN = Web3.utils.isBN
+export const isBN = BN.isBN
 // @ts-ignore
-export const isBigNum = Web3.utils.isBigNumber
```
rhlsthrm: 🙏

modules/client/src/controllers/RedeemController.test.ts
LayneHaber: if it is hex, but the wrong secret, it would fail on the hub

modules/client/src/controllers/RedeemController.ts
LayneHaber: i didnt think it was too valuable to add that case when were testing with mocked hubs

modules/hub/src/PaymentsService.ts
```diff
@@ -186,49 +186,63 @@ export default class PaymentsService {
...
-            recipient: 'hub',
-            // NOTE: args to user are generated in 
-            // `doInstantCustodialPayment`, this is a small hack
-            // to allow the hub to forward the payment
-          },
-          txCount: null,
-        }
+
+    // always check for collateralization regardless of payment status
+    const [res, err] = await maybe(this.channelsService.doCollateralizeIfNecessary(user))
```
rhlsthrm: This layout looks so much cleaner than the old one.

modules/hub/src/api/PaymentsApiService.ts
```diff
@@ -147,12 +147,15 @@ export class PaymentsApiServiceHandler {
...
+    } 
 
     const result = await this.paymentsService.doRedeem(user, secret)
     if (result.error != false) {
       LOG.warn(result.msg)
-      return res.send(400).json(result.msg)
+      // @ts-ignore
+      // TODO: wtf? it works, but doesnt compile
+      // are the express types out of date somehow?
+      return res.send(500, result.msg)
```
rhlsthrm: We should throw 400 or something different than 500 here, since it's not an uncaught exception.

rhlsthrm: https://expressjs.com/en/4x/api.html#res.status

See above for documented way to do what you're doing.

rhlsthrm: We should add this everywhere lol.

PR #118, opened 2019-03-07
BigNumber, intcomma fixes from SpankChain

wolever:


modules/client/src/controllers/BuyController.ts
```diff
@@ -69,6 +69,7 @@ export default class BuyController extends AbstractController {
           // PUNT on this -- AB
           break
         case 'PT_CHANNEL':
+        case 'PT_CUSTODIAL':
```
LayneHaber: So my guess is this won't cause any material problems, but it should be noted this type isn't implemented hub-side with this PR

modules/client/src/register/bignumber.ts
```diff
@@ -0,0 +1,34 @@
+import {BigNumber} from "bignumber.js"
+
+const oldBigNumberToString = BigNumber.prototype.toString
+BigNumber.prototype.toString = function(base?: number) {
+  // BigNumber.toString will only use exponential notation if a base is not
+  // specified. In dev, throw an error so it can be replaced with `.toFixed()`,
+  // or in production, log an error and force base 10.
+  if (!base) {
+    const err = new Error('Potentially unsafe of BigNumber.toString! Use .toFixed() instead.')
```
LayneHaber: 🥇

modules/client/src/types.ts
```diff
@@ -945,7 +945,7 @@ Returns:
 
 */
 
-export type PurchasePaymentType = 'PT_CHANNEL' | 'PT_THREAD' | 'PT_LINK'
+export type PurchasePaymentType = 'PT_CHANNEL' | 'PT_THREAD' | 'PT_CUSTODIAL' | 'PT_LINK'
```
LayneHaber: same note as above re: hub types

modules/client/src/controllers/BuyController.ts
rhlsthrm: Yep, should just fail on the hub immediately.

PR #142, opened 2019-03-18
Retry chainsaw on failures

rhlsthrm:
Chainsaw will now retry processing 10 times before shutting channel down.

### DIFF
--- a/modules/hub/src/ChainsawService.test.ts
+++ b/modules/hub/src/ChainsawService.test.ts
@@ -49,8 +49,6 @@ describe('ChainsawService::mocked Web3', function() {
   beforeEach(async () => {
     // ensure that fetchEvents will not return any new events
     // by making the top block the same as the last block on event (1)
-    if (registry)
-      await registry.clearDatabase()
 
     registry = getTestRegistry({
       Web3: {
@@ -67,6 +65,7 @@ describe('ChainsawService::mocked Web3', function() {
         _address: CM_ADDRESS,
       }
     })
+    await registry.clearDatabase()
     chan1 = await channelUpdateFactory(registry)
     chainsawDao = registry.get('ChainsawDao')
     cs = registry.get('ChainsawService')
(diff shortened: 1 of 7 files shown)


modules/hub/src/ChainsawService.test.ts
```diff
@@ -154,13 +153,22 @@ describe('ChainsawService::mocked Web3', function() {
   })
 
   describe('processSingleTx', () => {
-    it('should return poll type "SKIP_EVENTS" if state generation fails', async () => {
+    it('should return poll type "RETRY" if state generation fails', async () => {
      const pollType = await cs.processSingleTx(failedTxHash)
-     assert.equal(pollType, 'SKIP_EVENTS')
+     assert.equal(pollType, 'RETRY')
```
LayneHaber: all of the other poll types have the suffix `_EVENTS`, so i think we should add it here for consistency

modules/hub/src/ChainsawService.ts
```diff
@@ -206,39 +213,56 @@ export default class ChainsawService {
...
+      const sigHub = await this.signerService.signMessage(hash);
+
+      await this.channelsDao.applyUpdateByUser(event.user, 'ConfirmPending', this.config.hotWalletAddress, {
+        ...state,
+        sigHub
+      } as ChannelState, { transactionHash: event.txHash } as ConfirmPendingArgs, chainsawId)
+
+      return 'PROCESS_EVENTS'
     } catch (e) {
+      const NUM_RETRY_ATTEMPTS = 10
```
LayneHaber: would we ever want to make this configurable?

modules/hub/src/dao/ChainsawDao.ts
```diff
@@ -13,7 +13,7 @@ const STATUS_TO_STATES = {
   CS_SETTLED: 2,
 } as { [k: string]: number }
 
-export type PollType = 'FETCH_EVENTS' | 'PROCESS_EVENTS' | 'SKIP_EVENTS'
+export type PollType = 'FETCH_EVENTS' | 'PROCESS_EVENTS' | 'SKIP_EVENTS' | 'RETRY'
```
LayneHaber: should just add the suffix `_EVENTS` here for consistency

PR #147, opened 2019-03-21
Hotfix

rhlsthrm:
Pretty straightforward, will be testing manually in parallel to reviews.

### DIFF
--- a/modules/hub/src/ChainsawService.ts
+++ b/modules/hub/src/ChainsawService.ts
@@ -268,64 +268,70 @@ export default class ChainsawService {
   private async processDidStartExitChannel(chainsawId: number, event: DidStartExitChannelEvent) {
     const onchainChannel = await this.signerService.getChannelDetails(event.user)
 
-    const disputeRecord = await this.channelDisputesDao.getActive(event.user)
+    let disputeRecord = await this.channelDisputesDao.getActive(event.user)
     if (!disputeRecord) {
       // dispute might not have been initiated by us, so we need to add it here
-      await this.channelDisputesDao.create(event.user, 'Dispute caught by chainsaw', chainsawId, null, onchainChannel.channelClosingTime)
+      disputeRecord = await this.channelDisputesDao.create(event.user, 'Dispute caught by chainsaw', chainsawId, null, onchainChannel.channelClosingTime)
     } else {
       await this.channelDisputesDao.setExitEvent(disputeRecord.id, chainsawId, onchainChannel.channelClosingTime)
     }
 
     // check if sender was user
-    if (event.senderIdx == 1) {
-      const latestUpdate = await this.channelsDao.getLatestExitableState(event.user)
-      if (event.txCountGlobal <= latestUpdate.state.txCountGlobal) {
-        LOG.info(`Channel has not exited with the latest state, hub will respond! event ${prettySafeJson(event)}`)
-        const data = this.contract.methods.emptyChannelWithChallenge(
-          [latestUpdate.state.user, latestUpdate.state.recipient],
-          [
-            latestUpdate.state.balanceWeiHub.toFixed(),
-            latestUpdate.state.balanceWeiUser.toFixed()
-          ],
-          [
-            latestUpdate.state.balanceTokenHub.toFixed(),
-            latestUpdate.state.balanceTokenUser.toFixed()
-          ],
-          [
-            latestUpdate.state.pendingDepositWeiHub.toFixed(),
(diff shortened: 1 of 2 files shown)


modules/hub/src/ChainsawService.ts
```diff
@@ -272,10 +272,60 @@ export default class ChainsawService {
...
-      return
+    } else {
+      await this.channelDisputesDao.setExitEvent(disputeRecord.id, chainsawId, onchainChannel.channelClosingTime)
     }
 
-    await this.channelDisputesDao.setExitEvent(disputeRecord.id, chainsawId, onchainChannel.channelClosingTime)
+    // check if sender was user
+    if (event.senderIdx == 1) {
+      const latestUpdate = await this.channelsDao.getLatestExitableState(event.user)
+      if (event.txCountGlobal <= latestUpdate.state.txCountGlobal) {
```
wolever: This should be something like:

```
if (event.txCountGlobal < lastUpdate.txCountGlobal) {
  emptyChannelWithChallenge(…)
} else {
  emptyChannel()
}
```

modules/hub/src/ChainsawService.ts
```diff
@@ -272,10 +272,60 @@ export default class ChainsawService {
     if (!disputeRecord) {
       // dispute might not have been initiated by us, so we need to add it here
       await this.channelDisputesDao.create(event.user, 'Dispute caught by chainsaw', chainsawId, null, onchainChannel.channelClosingTime)
-      return
+    } else {
+      await this.channelDisputesDao.setExitEvent(disputeRecord.id, chainsawId, onchainChannel.channelClosingTime)
```
wolever: I'm not 100% certain, but it seems like this should be:

```
if (!disputeRecord)
  disputeRecord = this.channelDipsutesDao.create(…)
await this.channelDisputesDao.setExitEvent(…)
```

modules/hub/src/ChainsawService.ts
```diff
@@ -272,10 +272,60 @@ export default class ChainsawService {
...
     }
 
-    await this.channelDisputesDao.setExitEvent(disputeRecord.id, chainsawId, onchainChannel.channelClosingTime)
+    // check if sender was user
+    if (event.senderIdx == 1) {
```
wolever: For readability, I'd have:

```
if (event.senderIdx == 0)
  return

… exit channel …
```

modules/hub/src/ChainsawService.ts
rhlsthrm: Absolutely correct.

modules/hub/src/ChainsawService.ts
```diff
@@ -268,14 +268,70 @@ export default class ChainsawService {
...
+      LOG.info(`Hub inititated the challenge, so no need to respond; event ${prettySafeJson(event)}`)
       return
     }
 
-    await this.channelDisputesDao.setExitEvent(disputeRecord.id, chainsawId, onchainChannel.channelClosingTime)
+    let txn
+    let data
+    const latestUpdate = await this.channelsDao.getLatestExitableState(event.user)
+    if (event.txCountGlobal <= latestUpdate.state.txCountGlobal) {
+      LOG.info(`Channel has not exited with the latest state, hub will respond! event ${prettySafeJson(event)}`)
```
wolever: Should probs also have a log message that we're emptying channel, along with the event and the latest channel event.

modules/hub/src/ChainsawService.ts
```diff
@@ -268,14 +268,70 @@ export default class ChainsawService {
...
+    }
+
+    // check if sender was user
+    if (event.senderIdx == 0) {
+      LOG.info(`Hub inititated the challenge, so no need to respond; event ${prettySafeJson(event)}`)
       return
     }
 
-    await this.channelDisputesDao.setExitEvent(disputeRecord.id, chainsawId, onchainChannel.channelClosingTime)
+    let txn
```
wolever: This can be defined down at the bottom where it's assigned

modules/hub/src/ChainsawService.ts
wolever: Should there still be the `} else { this.channelDisputesDao.setExitEvent(…) }` here?

ArjunBhuptani: Aren't those the same thing?

rhlsthrm: @ArjunBhuptani, no, one is creating the record in the db, one is modifying an existing record in the db.

PR #149, opened 2019-03-22
Sc connext hub do not merge

rhlsthrm:
Tests are passing, can we merge?


modules/hub/README.rst
```diff
@@ -26,6 +26,14 @@ Running Tests
 
    $ ./migrate-and-test.sh
 
+Creating Migrations
```
LayneHaber: add here that you have to add it to the database module as well. we could make a precommit hook for that?

modules/hub/infrastructure/helm-values-production.yaml
```diff
@@ -28,6 +28,18 @@ env:
   SERVICE_USER_KEY: 'nnWZrKw7WuE0K1JWb2PHup42xEVrLDnkMTcEbqecF3NeuLcleGl7crJwo2Hc0Uh'
   CHANNEL_MANAGER_ADDRESS: '0xbf2aeab03b82c182599930217272ef0fc5cf662b'
   TOKEN_CONTRACT_ADDRESS: '0x6b01c3170ae1efebee1a3159172cb3f7a5ecf9e5'
+  SHOULD_COLLATERALIZE_URL: 'https://beta.spankchain.com/api/model-signup/should-collateralize/'
```
LayneHaber: do we want to commit these?

wolever: Can definitely be removed

wolever: (and should also be removed on our end)

rhlsthrm: @wolever, is this supposed to be ignored?

modules/hub/README.rst
rhlsthrm: @bohendo how do you think we should handle this? We should probably have the CI process pull in migrations from the hub module into the db module.

rhlsthrm:
Opening this PR against staging.

PR #153, opened 2019-03-26
Onchain tx rearchitecture

rhlsthrm:
## Types of changes

## Checklist:


modules/hub/src/dao/OnchainTransactionsDao.ts
```diff
@@ -65,6 +66,21 @@ export class OnchainTransactionsDao {
...
+  async getLatestConfirmed(db: DBEngine, account: string): Promise<OnchainTransactionRow | null> {
+    const row = await db.queryOne(SQL`
+      SELECT *
+      FROM onchain_transactions_raw
+      WHERE
+        "from" = ${account} AND
+        state = 'confirmed'
+      ORDER BY nonce DESC
+      LIMIT 1
+      FOR UPDATE
```
wolever: The `FOR UPDATE` is unnecessary here, since this row isn't going to be updated.

modules/hub/src/OnchainTransactionService.ts
```diff
@@ -333,6 +332,53 @@ export class OnchainTransactionService {
...
+        // TODO: what errors can happen here?
+        LOG.warning(`Error checking status of tx '${txn.hash}': ${'' + err} (will retry)`)
+        return
+      }
+
+      if (tx) {
+        // if tx exists at this point, mark as submitted
+        await this.updateTxState(txn, {
+          state: 'submitted',
+        })
```
wolever: There should be a return after this, since we should never check timeouts etc if we've seen the tx on chain.

modules/hub/src/OnchainTransactionService.ts
```diff
@@ -301,10 +268,26 @@ export class OnchainTransactionService {
...
 
-      if (!tx || !tx.blockNumber) {
-        const txnAgeS = (Date.now() - (+new Date(txn.submittedOn))) / 1000
-        const txnAge = `${Math.floor(txnAgeS / 60)}m ${Math.floor(txnAgeS % 60)}s`
+      const txnAgeS = (Date.now() - (+new Date(txn.submittedOn))) / 1000
+      const txnAge = `${Math.floor(txnAgeS / 60)}m ${Math.floor(txnAgeS % 60)}s`
+
+      // not in mempool yet
+      if (!tx) {
+        // resubmit after 60 seconds of waiting
```
wolever: update these comments

rhlsthrm:
Fixed according to your comments @wolever.

PR #154, opened 2019-03-26
Quiet logs + fix dashboard routing

bohendo:
## Description

No more logging full channel state when the client syncs. We're still logging the channel state on important events like when the user deposits though.

Dashboard routing is improved & the mainnet hub url isn't hardcoded into the client bundle anymore so the same bundle should work on both rinkeby & mainnet.

## How Has This Been Tested?

CI tests are green & e2e tests pass in the card.

## Types of changes

## Checklist:


modules/dashboard/src/components/ContractInfoCard.js
```diff
@@ -17,35 +17,35 @@ const styles = theme => ({
...
+          <>
+            <Typography className={classes.pos} color="textSecondary">
+              <a href={`https://etherscan.io/address/${contractAddress}`} target="_blank" rel="noopener noreferrer">{contractAddress}</a>
+            </Typography>
+            <Typography variant="h5" component="h2">
+              {parseFloat(wei.formatted).toFixed(2)}... ETH ({wei.raw} Wei)
+            </Typography>
             <Typography variant="h5" component="h2">
-              Loading...
+              ${parseFloat(token.formatted).toFixed(2)}... DAI ({token.raw} Dei)
```
bohendo: I didn't make this change, there must have been some updates to master that didn't go through staging..?

modules/dashboard/src/components/Dashboard.js
```diff
@@ -199,12 +191,12 @@ class Dashboard extends React.Component {
                 </IconButton>
               </div>
               <Divider />
-              <List>{mainListItems}</List>
+              <SidebarLinks publicUrl={publicUrl} />
```
bohendo: Replaced the `mainListItems` constant w a react component function called `SidebarLinks` so I could pass in the url as a prop.

modules/dashboard/src/components/Deposits.js
```diff
@@ -96,45 +96,45 @@ class Deposits extends Component {
...
-          tickValues={[0, 1, 2, 3, 4, 5, 6, 7]}
-          label="Day"
-          style={{ axisLabel: { fontSize: 2 }, tickLabels: { fontSize: 2 } }}
-        />
-  </VictoryChart> 
-    ); 
-
-    console.log(toRender);
-    return toRender
+      const toRender = (
```
bohendo: Cleaned up some trailing whitespace & improper indentation.

modules/dashboard/src/index.js
```diff
@@ -5,13 +5,18 @@ import * as serviceWorker from './serviceWorker';
...
+ReactDOM.render(
+  <Index publicUrl={urlPrefix} ethUrl={ethUrl} hubUrl={hubUrl} apiUrl={apiUrl}/>,
+  document.getElementById('root')
+);
 
 // If you want your app to work offline and load faster, you can change
 // unregister() to register() below. Note this comes with some pitfalls.
 // Learn more about service workers: http://bit.ly/CRA-PWA
-serviceWorker.unregister();
+serviceWorker.register();
```
bohendo: I updated the serviceWorker to use `window.location.origin` instead of `process.env.PUBLIC_URL` so it'll work on both rinkeby & mainnet. Tentatively turned it on during testing to make sure that my changes didn't break anything but I don't think this is necessary. We can keep the serviceWorked deactivated if it ends up causing problems.

modules/hub/src/ChannelsService.ts
```diff
@@ -106,10 +106,8 @@ export default class ChannelsService {
     }, user)
 
     if (hasPendingOps(channelStateStr)) {
-      LOG.info(
-        `User requested a deposit while state already has pending operations ` +
-        `(user: ${user}; current state: ${JSON.stringify(channelStateStr)})`
-      )
+      LOG.info(`User ${user} requested a deposit while state already has pending operations `)
+      LOG.debug(`[Request Deposit] Current state: ${JSON.stringify(channelStateStr)}`)
```
bohendo: I kept similar hub logs in `info` log-level minus the channel state. I didn't remove the channel state logging but switched it to only print on `debug` log-level. The default log-level is `info` so these won't be printed unless we explicitly start the hub in `debug` log-level-mode.

modules/hub/src/ChannelsService.ts
```diff
@@ -335,10 +335,11 @@ export default class ChannelsService {
   ): Promise<WithdrawalArgs | null> {
     const channel = await this.channelsDao.getChannelByUser(user)
     if (!channel || channel.status !== 'CS_OPEN') {
-      throw new Error(
+      LOG.error(
```
bohendo: In a couple cases, normal operation will result in the hub throwing an error & printing out a bug ugly stack track. I replaced this with a `LOG.error` followed by a `return`. I assume this'll work the same but be a little quieter.

PR #172, opened 2019-04-02
0 timeout invalidations

LayneHaber:
## Description
Overhaul of invalidation mechanism

## Motivation and Context
Invalidation mechanism should invalidate the pending states by unwinding pending ops, not by invalidating a range of states, so you can use operating balance on 0 timeout states.

## How Has This Been Tested?
Tested invalidating withdrawals, user deposits, and hub collateralizing invalidations while making payments with operating balance locally. 

Added additional tests into the client + hub.

## Types of changes


modules/client/src/StateGenerator.ts
```diff
@@ -74,6 +75,18 @@ export function calculateExchange(args: ExchangeArgsBN) {
   }
 }
 
+export function depositIfNegative(r: any, src: string, dst: string) {
```
rhlsthrm: Can we change `any` to something like `{ [key: string]: string }` or something similar?

modules/client/src/testing/index.test.ts
```diff
@@ -59,6 +59,24 @@ describe('expandSuccinctChannel', () => {
       },
     )
   })
+
+  it('should definitely work on my mama', () => {
```
rhlsthrm: ???

rhlsthrm: lol

rhlsthrm:
Re-open against master, we don't have staging anymore.

PR #195, opened 2019-04-16
[WIP] 187 optimistic payments

LayneHaber:
Payments are queued for processing and held by hub until user comes online if specified

## The Problem
Queueing payments on hub side rather than client/implementer side

## The Solution
Added payments_optimistic table, that polls the recipient's channel balance and makes the payment if needed

## Checklist: 

TO BE COMPLETED AFTER [WIP] LABEL REMOVED


modules/hub/src/OptimisticPaymentsService.ts
```diff
@@ -0,0 +1,114 @@
...
+      const redemptionId = await this.paymentsService.doChannelInstantPayment(purchasePayment, payment.paymentId, payment.channelUpdateId)
+
+      await this.opPaymentDao.addOptimisticPaymentRedemption(payment.paymentId, redemptionId)
+    } catch (e) {
+      // if the custodial payment fails, the payment should fail
+      LOG.info("Error redeeming optimistic channel payment. ID: {id}", {
+        id: payment.paymentId,
+      })
+    }
+    // TODO: recollateralization here?
```
rhlsthrm: We should trigger a collateralization check before the payment attempt, I believe we do that for the other payments, right?

modules/hub/src/dao/ChannelsDao.ts
```diff
@@ -269,10 +279,19 @@ export class PostgresChannelsDao implements ChannelsDao {
...
         WHERE
           recipient = ${user} AND
           created_on > NOW() - ${this.config.recentPaymentsInterval}::interval
+
+        UNION
+
+        SELECT * FROM payments
+        WHERE 
+          recipient = ${user} AND
+          payment_type = 'PT_OPTIMISTIC'
```
rhlsthrm: Should we also make sure status != "FAILED"?

modules/hub/src/domain/OptimisticPayment.ts
```diff
@@ -0,0 +1,13 @@
+import { Omit } from "../vendor/connext/types";
+import { PurchasePaymentRow } from "./Purchase";
+
+
+export type OptimisticPaymentStatus = "NEW" | "COMPLETED" | "FAILED"
+
+export type OptimisticPurchasePaymentRow = Omit<PurchasePaymentRow, "type" | "id" | "custodianAddress"> & {
```
rhlsthrm: Does it make more sense to put these in the client types?

modules/hub/src/services.ts
```diff
@@ -400,15 +410,36 @@ export const serviceDefinitions: PartialServiceDefinitions = {
...
       'CustodialPaymentsDao',
       'Validator',
       'Config',
       'DBEngine',
-      'GlobalSettingsDao',
+      'GlobalSettingsDao'
     ],
   },
 
+  OptimisticPaymentsService: {
```
rhlsthrm: I think the poller service will need to be a singleton.

modules/hub/src/OptimisticPaymentsService.ts
```diff
@@ -0,0 +1,114 @@
...
+import ChannelsDao from "./dao/ChannelsDao";
+import DBEngine from "./DBEngine";
+import { convertChannelState, convertPayment, PurchasePayment } from "./vendor/connext/types";
+import OptimisticPaymentDao from "./dao/OptimisticPaymentDao";
+import { OptimisticPurchasePaymentRow } from "./domain/OptimisticPayment";
+import PaymentsService from "./PaymentsService";
+
+const LOG = log('OptimisticPaymentsService')
+
+const POLL_INTERVAL = 1000
```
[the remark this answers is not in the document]
LayneHaber: it affects the speed at which payments are resolved once collateral is in, so it would be a potential poller length interval of wait time on the ui, can turn it down though

LayneHaber:
merged in with connext-api see #197

PR #197, opened 2019-04-17
Connext API

bohendo:
## The Problem

Integrating a connext client was impossible w/out copying our Frankenstein's monster of a web3 instance from the dai card. Previous teams resorted to wrapping our lib in their own wrapper w a more sane API.

This PR aims to save future teams the trouble. See [specs](https://paper.dropbox.com/doc/Connext-API--Abb3grRtFyseLqlKNojt~9YgAg-yLiQGDEsN3hYkYdmSNSP7) for details.

## The Solution

Implement a Wallet class in the client. This wallet is our client's master signer & all access to the private key goes through it. This gives us the flexibility to sign w a custom-made web3 if provided or a mnemonic or private key or etc etc all these decisions regarding how to sign now happen in a single place.

See the card's [Connext API PR](https://github.com/ConnextProject/card/pull/38) for an example of how the the new API can be used.

These changes are entirely backwards compatible though so the daicard will still work as-is if it's client is upgraded to include these changes.

## Checklist:

Documentation updates coming soon (eg Monday probs)

TODO: publish connext to npm (version already incremented) once these changes are good to go.


modules/client/src/Wallet.ts
```diff
@@ -0,0 +1,119 @@
+import { ethers as eth } from 'ethers';
+import Web3 from 'web3';
+import { ConnextClientOptions } from './Connext';
+import { Provider } from './types'
+
+export default class Wallet extends eth.Signer {
```
LayneHaber: I have a feeling this class (if it's used by others) will need a more specific name, something like `ConnextWallet` or `SigningWallet` or something.

If wallets integrate this package, the nomenclature here may be confusing

LayneHaber: Also, when we merge this let's make sure to have a correspondingly named PR in the card (and probably docs) that implement these updates

rhlsthrm: But I think it'll be `Connext.Wallet` right? Or `import { Wallet } from 'Connext'`, they can always say `import { Wallet as ConnextWallet } from 'Connext'` if they want to avoid overriding.

modules/client/src/Wallet.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  async signTransaction(tx: any) {
+    tx.to = await tx.to // Why is this sometimes a promise?!
+    if (this.signer) {
+      return await this.signer.sign(tx)
+    }
+    if (this.web3) {
+      return await (
+        this.web3.eth.signTransaction
```
rhlsthrm: Same as above, not sure why we are checking the existence of the function here.

modules/client/src/validator.ts
```diff
@@ -67,14 +75,15 @@ export class Validator {
 
   private generateHandlers: { [name in ChannelUpdateReason]: any }
 
-  web3: any
-
+  provider: any
+  abi: any
```
rhlsthrm: Can we make this a type?

modules/client/src/Hub.ts
```diff
@@ -0,0 +1,393 @@
+import * as eth from 'ethers';
```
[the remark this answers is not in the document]
bohendo: Last time we tried to mix the two, we didn't have e2e tests so had trouble noticing issues before they were in prod, we're better protected now & can make sure the usages are properly separated before deploying.

modules/client/package.json
```diff
@@ -35,12 +35,13 @@
     "dotenv": "^6.2.0",
     "ethereumjs-util": "^6.0.0",
     "ethereumjs-wallet": "0.6.0",
+    "ethers": "^4.0.27",
```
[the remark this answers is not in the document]
bohendo: Heh well web3.js depends on ethers so it's going to end up in `node_modules` either way. But yea, this does make the `toString()` stuff a non-issue. More importantly, ethers is more flexible and makes it simple to incorporate several different signing methods eg a private key, mnemonic, or web3 signer.

modules/client/src/Wallet.ts
bohendo: No no no this `Wallet` class is only used internally, there shouldn't be any reason that the daicard or metamask or anyone needs to access this wallet class directly. It's attached to the `ConnextInternal` instance but I see no reason to expose this on the `ConnextClient` instance that the end user interacts with.

modules/client/src/Wallet.ts
```diff
@@ -0,0 +1,118 @@
...
+    return this.address
+  }
+
+  async signMessage(message: string) {
+    if (this.signer) {
+      return await this.signer.signMessage(eth.utils.arrayify(message))
+    }
+    if (this.web3) {
+      return await (
+        this.web3.eth.sign
```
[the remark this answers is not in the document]
bohendo: Good call

modules/client/src/StateGenerator.ts
```diff
@@ -1,29 +1,29 @@
...
+  ThreadState,
+  ThreadStateBN,
   UpdateRequestBN,
-  InvalidationArgs,
+  UnsignedChannelState,
+  UnsignedChannelStateBN,
+  UnsignedThreadState,
+  UnsignedThreadStateBN,
   VerboseChannelEventBN,
+  WithdrawalArgsBN,
```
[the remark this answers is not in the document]
bohendo: Fingers, mostly
Reviewing what depends on what helps me ramp up a new codebase, Alphabeticalization keeps my fingers busy as I read & makes things easier to understand at a glance..
But if you know any tools that'll do this automatically I'm all ears lol

modules/hub/src/ChannelsService.ts
```diff
@@ -1,56 +1,54 @@
...
 import { SignerService } from './SignerService';
 import { OnchainTransactionRow } from './domain/OnchainTransaction';
 import ChannelDisputesDao from './dao/ChannelDisputesDao';
-import { StateGenerator } from './vendor/connext/StateGenerator';
 import { CoinPaymentsDao } from './coinpayments/CoinPaymentsDao'
 import { OnchainTransactionsDao } from './dao/OnchainTransactionsDao';
-
+import { BigNumber as BN } from 'ethers/utils'
+import { assetToWei } from 'connext/dist/lib/bn';
+import { ethers } from 'connext/node_modules/ethers';
```
bohendo: Why aren't we just importing `from 'ethers'`? This breaks the build on my computer + on CI

modules/hub/src/ChannelsService.ts
```diff
@@ -1,56 +1,54 @@
...
 import ThreadsService from './ThreadsService';
 import { SignerService } from './SignerService';
 import { OnchainTransactionRow } from './domain/OnchainTransaction';
 import ChannelDisputesDao from './dao/ChannelDisputesDao';
-import { StateGenerator } from './vendor/connext/StateGenerator';
 import { CoinPaymentsDao } from './coinpayments/CoinPaymentsDao'
 import { OnchainTransactionsDao } from './dao/OnchainTransactionsDao';
-
+import { BigNumber as BN } from 'ethers/utils'
+import { assetToWei } from 'connext/dist/lib/bn';
```
bohendo: I'd really really like us to expose everything we need from connext via it's index. Importing individual files will lock up our file structure and make it more likely that future refactors break stuff unexpectedly.
We need to spend some time finalizing the connext API specs. We need to decide & make explicit which utilities are important to expose & which are internal libs that apps shouldn't need to mess with. If assetToWei is useful to apps that integrate connext, it should be available via the Utils or similar.

rhlsthrm: Why are we importing from `dist/` not `index/`? This is really strange to do.

LayneHaber: Weird, seems to be an issue with the autoimports in my code editor. Fixed it here, and in a couple other files as well.

The reason I changed the export structure was because I added a `hubAddress` to the `Utils` class. This allows the recover functions within the `Utils` class to have the same API as the `Validator` class.

We can change it back to a static class, and export everything through there, or we can change the exports to export a better `utils` object that includes everything. But, I do think we should rename the `Utils` class to `StateUtils` or something

PR #200, opened 2019-05-03
Tx history

rhlsthrm:
## Checklist:

### DIFF
--- a/modules/client/src/Connext.ts
+++ b/modules/client/src/Connext.ts
@@ -36,6 +36,7 @@ import {
   UnsignedChannelState,
   WithdrawalParameters,
   PartialPurchaseRequest,
+  PurchasePaymentRow,
 } from './types';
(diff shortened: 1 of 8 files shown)


modules/client/src/Connext.ts
```diff
@@ -166,6 +167,10 @@ export abstract class ConnextClient extends EventEmitter {
   async redeem(secret: string): Promise<{ purchaseId: string }> {
     return await this.internal.redeemController.redeem(secret)
   }
+
+  async getPaymentHistory(): Promise<PurchasePaymentRow[]> {
```
LayneHaber: need to add the `getPaymentHistoryById` to the external class otherwise looks good to me

LayneHaber: need to add the getPaymentHistoryById to the external class otherwise looks good to me

PR #201, opened 2019-05-03
fix: reassign `this.signer` when connect is called, and set gasPrice and nonce

dankins:
## The Problem

I am initializing Connext with a privateKey string, first using ganache-cli, then using Rinkeby Infura RPC.
On Ganache, I kept encountering "invalid nonce" errors, so then I gave up and thought it was a ganache issue, so I tried Rinkeby. On Rinkeby I was seeing "transaction underpriced" errors. 

## The Solution

After setting the gas price it fixed the "transaction underpriced" issues, but then I received nonce issues on Rinkeby. I then noticed that the connext Wallet class is overriding the Signer, so I looked at the implementation of Wallet in ethers and noticed it was setting the nonce there, but not in the Connext inherited version.
https://github.com/ethers-io/ethers.js/blob/master/src.ts/wallet.ts#L95

Adding the nonce fixed the Rinkeby issues, and I presume would have fixed my ganache issues as well. 

I have only been testing the Deposit function, so I presume other methods are still broken.

This introduces a backwards incompatible change where the wallet will throw if you do not provide either `web3`, `privateKey` or `mnemonic`. Happy to yank this out if you like. 

## Checklist:

### DIFF
--- a/modules/client/src/Wallet.ts
+++ b/modules/client/src/Wallet.ts
@@ -42,13 +42,13 @@ export default class Wallet extends eth.Signer {
     // First choice: Sign w private key
     if (opts.privateKey) {
       this.signer = new eth.Wallet(opts.privateKey || '')
-      this.signer.connect(this.provider)
+      this.signer = this.signer.connect(this.provider)
       this.address = this.signer.address.toLowerCase()
 
     // Second choice: Sign w mnemonic
     } else if (opts.mnemonic) {
       this.signer = eth.Wallet.fromMnemonic(opts.mnemonic || '')
-      this.signer.connect(this.provider)
+      this.signer = this.signer.connect(this.provider)
       this.address = this.signer.address.toLowerCase()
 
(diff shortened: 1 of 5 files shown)


modules/client/src/Wallet.ts
```diff
@@ -59,10 +59,7 @@ export default class Wallet extends eth.Signer {
 
     // Default: create new random mnemonic
     } else {
-      this.signer = eth.Wallet.createRandom()
-      this.signer.connect(this.provider)
-      this.address = this.signer.address.toLowerCase()
-      console.warn(`Generated a new signing key, make sure you back it up before sending funds`)
+      throw new Error("please provide either `privateKey`, `mnemonic`, or `web3` in the ConnextOptions")
```
rhlsthrm: Can we still do the random mnemonic creation here?

LayneHaber:
Closed in favor of #202 to properly run changes through CI

PR #203, opened 2019-05-06
Fixes for contract types

rhlsthrm:
## The Problem
Web3 now returns BigNumber types for all the number fields in contract event logs. The current chainsaw code assumes that numbers are returned and directly inserts into the DB as a serialized object.

## The Solution
Changed the contract event poller/transformer to convert the returned values to numbers or strings before inserting into the DB. Had to use a lot of `any` casts which I hate, but otherwise the whole system needs to be changed, which we should do as a bigger overhaul when we do things like switch web3 for ethers.

## Checklist:


modules/hub/src/domain/ContractEvent.ts
```diff
@@ -189,7 +189,7 @@ export class DidUpdateChannelEvent extends ContractEvent {
   toFields (): Object | any {
     return {
       user: this.user.toLowerCase(),
-      senderIdx: this.senderIdx,
+      senderIdx: (this.senderIdx as any).toNumber(),
```
LayneHaber: Looks good to me, may want add a note about how this will fail if its outside of the safe range but otherwise nicely done

rhlsthrm: SenderIdx can only be 0 or 1, and we cast the txCounts as number anyways.

PR #207, opened 2019-05-10
Payment profiles

LayneHaber:
## The Problem
Need a way to set expected collateral needs for a group of user addresses.

## The Solution
Added the MVPaymentProfile. Channels can be linked to a `payment_profile_id` by an admin, and the parameters are stored in a new `payment_profiles` table.

Key parts of the code will be migration and dao related, but it is an admin only endpoint at the moment so we should be okay.

## Checklist:


modules/database/migrations/sqls/20190507194131-add-payment-profiles-up.sql
```diff
@@ -0,0 +1,16 @@
+
+/* Simple payment profile with min collateral and collateral target */
+create table payment_profiles (
+	id bigserial primary key,
+  minimum_maintained_collateral_wei wei_amount default '0',
```
rhlsthrm: Minor nitpick, if you're going to default it like this, you might as well make it `not null`.

modules/database/migrations/sqls/20190507194131-add-payment-profiles-up.sql
```diff
@@ -0,0 +1,16 @@
...
+  amount_to_collateralize_wei wei_amount default '0',
+  amount_to_collateralize_token wei_amount default '0'
+);
+
+/* Add payment profile id to the channels column */
+alter table _cm_channels add column payment_profile_id bigint null;
+alter table _cm_channels add constraint payment_profile_id_fk foreign key (payment_profile_id) references payment_profiles(id);
+
+/* Update the channels view */
+create or replace view cm_channels as (select * from _cm_channels);
```
rhlsthrm: We probably honestly shouldn't be doing this, we should only put what we need/use in the view but let's worry about that later.

modules/hub/src/PaymentProfilesService.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+const {
+  toWeiString,
+} = big
+
+describe('PaymentProfilesService', () => {
+  const registry = getTestRegistry()
+  const service = registry.get('PaymentProfilesService')
+
+  // **** helper functions
```
rhlsthrm: Very nice.

modules/hub/src/api/PaymentProfilesApiService.ts
```diff
@@ -0,0 +1,149 @@
...
+    if (!config) {
+      res.status(400)
+      res.send({ error: `No payment profile config found with id: ${id}` })
+    }
+
+    return res.send(config)
+  }
+
+  async doGetPaymentProfileByUser(req: express.Request, res: express.Response) {
+    if (!isAdmin(req)) {
```
rhlsthrm: I think it's fine to make this `ownedAddressOrAdmin`, users should be able to see their own profile I think.

LayneHaber: Made it with user auth, since we specd it out client side

PR #211, opened 2019-05-17
Support for external wallets

brunobar79:
## The Problem

External wallets like MetaMask have their own accounts and signing mechanisms so passing private keys or mnemonics is not an option. 

## The Solution

Instead I propose a new signing strategy which involves passing an object to the client constructor with an identical interface that will take care of signing messages and proposing transactions
```
const opts = {
      ...,
      externalWallet: {
	external: true,
	address: this.myExternalWalletAddress,
	getAddress: () => Promise.resolve(this.myExternalWalletAddress),
	getBalance: block => provider.getBalance(this.myExternalWalletAddress, block),
	signMessage: message => myCustomFnToSignMessages(message)
	sign: tx => myCustomFnToSignAndSubmitTxs(message)
    }
}
```

## Checklist:


modules/client/src/Wallet.ts
```diff
@@ -5,7 +5,7 @@ import { IConnextChannelOptions } from './Connext'
 import {
   objMapPromise,
   TransactionRequest,
-  TransactionResponse,
+  TransactionResponse
```
bohendo: We thought KT's vote for trailing commas was the tie breaker but apparently this debate is tied up again 😅

modules/client/src/Wallet.ts
```diff
@@ -95,30 +102,47 @@ export class Wallet extends eth.Signer {
...
   }
-
-  public async signTransaction(tx: TransactionRequest): Promise<string> {
-    if (this.signer) {
-      return this.signer.sign(tx)
-    }
-    if (this.web3) {
-      // resolve any promise fields
+  
+  public async prepareTransaction(tx: TransactionRequest): Promise<any> {
```
bohendo: Could we make this private?

modules/client/src/Wallet.ts
```diff
@@ -95,30 +102,47 @@ export class Wallet extends eth.Signer {
...
+    }
+    if (this.web3) {
+      const txObj:any = await this.prepareTransaction(tx);
       return (await this.web3.eth.signTransaction(txObj)).raw // TODO: fix type
     }
     throw new Error(`Could not sign transaction`)
   }
+  
+  
+  public async signTransactionExternally(tx: TransactionRequest): Promise<any> {
```
bohendo: This one could probably be private too.

modules/client/src/Wallet.ts
```diff
@@ -52,8 +53,14 @@ export class Wallet extends eth.Signer {
       this.signer = eth.Wallet.fromMnemonic(opts.mnemonic || '')
       this.signer = this.signer.connect(this.provider)
       this.address = this.signer.address.toLowerCase()
-
-    // Third choice: Sign w web3
+    // Third choice: External wallets
+    } else if (opts.externalWallet) {
+      this.signer = opts.externalWallet
+      this.external = true
+      this.web3 = new Web3(opts.web3Provider as any)
```
bohendo: I'm a little confused, while using an external wallet:
 - If `signTransaction` is called, it'll ultimately be passed to `opts.externalWallet.sign()`
 - If `sendTransaction` is called, it'll end up at `web3.eth.sendTransaction()`
 - If `signMessage` is called, it'll end up at `opts.externalWallet.signMessage()`

Why use `externalWallet` interface to sign some stuff and `web3` in other cases? Could we just use one or the other?

modules/client/src/Wallet.ts
```diff
@@ -95,30 +102,47 @@ export class Wallet extends eth.Signer {
...
+    }
+    throw new Error(`Could not sign transaction`)
+  }
 
   public async sendTransaction(txReq: TransactionRequest): Promise<TransactionResponse> {
+    if (this.external){
+      return this.signTransactionExternally(txReq);
+    }
+
     txReq.gasPrice = await (txReq.gasPrice || this.provider.getGasPrice())
```
bohendo: The `prepareTransaction` is nice, should we set gasPrice etc there too?

brunobar79:
@bohendo Thanks for the review!  I think all the comments have been addressed.

The `externalWallet` object has an almost identical interface to the Signer interface from 'ethers` abstract signer.

It should be the responsibility of the wallet itself to wire up the two functions: `sign`  & `signMessage`
Normally wallets should trigger the send transaction UI while calling `sign` but they shouldn't trigger the UI at all while signing messages, because of the reasons that you already mentioned.
This is exactly what I did with MetaMask for now. Happy to add a gif to clarify.

LayneHaber:
CI won't run on forks, all tests are passing locally so I'll merge it.

PR #223, opened 2019-05-29
Money2020 revisions

LayneHaber:
## The Problem
No easy way to track duplicates in email from client

## The Solution
Made temporary email table. Can be removed after event.

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,6 +84,7 @@ jobs:
               export INDRA_DOMAINNAME="$MAINNET_URL"
               export INDRA_ETH_NETWORK="mainnet"
               export INDRA_MODE="live"
+              export MAILGUN_API_KEY="$MAILGUN_API_KEY"
               bash ops/restart.sh prod
             EOF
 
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -57,6 +57,7 @@ jobs:
               export INDRA_MODE="staging"
               export INDRA_ETH_NETWORK="rinkeby"
               export INDRA_DOMAINNAME="$STAGING_URL"
+              export MAILGUN_API_KEY="$MAILGUN_API_KEY"
```
rhlsthrm: We'll need this in the live deploy as well.

modules/hub/src/PaymentsService.ts
```diff
@@ -86,14 +102,23 @@ export default class PaymentsService {
...
     const emailData = {
       from: 'Connext <requests@money2020.connext.network>',
       to,
-      subject,
+      subject: prefixedSubject,
       text,
     }
     try {
       const res = await mg.messages().send(emailData)
+      await this.db.withTransaction(
```
rhlsthrm: There's only one DB query it looks like, does this need to be in a transaction?

PR #228, opened 2019-06-06
Fix external wallets

brunobar79:
Need to swap the order of these two, since an external wallet also has  `this.signer` and it will never reach that code.


rhlsthrm:
Closing in favor of #229, which is on our fork

PR #250, opened 2019-08-26
Bot farm

LayneHaber:
Make sure everything is started and test with:

`NUMBER_BOTS=<{some number >=4}> bash ops/test-bot-farm.sh`

Does not *fix* any issues we've seen on prod, but does reproduce them locally. Also contains a breaking change for the client api where the payment id and the preimage must be provided to the client conditional transfer call. See [here](https://github.com/ConnextProject/indra/blob/2861d5cbf887f49e28c9dd7195d1bd024ad95c30/modules/daicard/src/components/sendCard.js#L183)


bohendo:
Looks good except this PR should be made against staging, not master.

PR #306, opened 2019-09-09
Fix rounding error in daicard's currency conversions

### LINKED ISSUE #257: Payments are not the expected value (they lose dust)
**Describe the bug**

Payments (via send or send links) are a very small decimal value off of what they should be. E.g., User A sends 1 DAI and User B receives more like 0.9999999999999.

**To Reproduce**

Discovered during the Mosendo x Connext bug testing prior to ETH Berlin. If you guys need more details, I can try to reproduce and screenshot

**Expected behavior**

**Screenshots**

**Your Environment**

* Client version used:
* Node version used:
* Chain used: 
* Browser Name and version:
* Operating System and version (desktop or mobile):
* Link to your project:

**Additional context**

bohendo:
## The Problem

Originally noticed by mosendo team, sends appeared to have a value tainted by a rounding error. 

Closes #257 

## The Solution

Use 36 units of precision to do division instead of 18. This ensures that any rounding errors (unavoidable in cases of irrational numbers eg 4/9) are at much less significant sigfigs and that, with proper rounding, we'll get the correct numbers in the first 18 digits (eg all that matters when working with wei).

Oops and this change is also bundled in with renaming all the cfModule/cfNode stuff to cfCore. Should have separated these into two separate PRs, my bad.

## Checklist:


modules/daicard/src/utils/currency.js
```diff
@@ -168,19 +163,36 @@ export class Currency {
       this.daiRate = daiRate;
       this.daiRateGiven = true;
     }
-    const amountInWei = tokenToWei(this.wad, this.getExchangeRate(this.type))
-    const targetAmount = fromWei(weiToToken(amountInWei, this.getExchangeRate(targetType)))
+    const thisToTargetRate = this.toRay(this.getRate(targetType)).div(this.getRate(this.type))
+    const targetAmount = this.fromRay(this.fromRoundRay(this.ray.mul(thisToTargetRate)))
```
bohendo: New currency conversion math utilizes ray units

modules/daicard/src/utils/currency.js
```diff
@@ -168,19 +163,36 @@ export class Currency {
...
     return new Currency(
       targetType,
       targetAmount.toString(),
       this.daiRateGiven ? this.daiRate : undefined,
     )
   }
 
+  // convert to wad, add 0.5 wad, convert back to dec string, then truncate decimal
+  _round = (decStr) =>
+    this._floor(this.fromWad(this.toWad(decStr).add(this.toWad('0.5'))).toString())
```
bohendo: This fn handles rounding eg when converting back to wei from ray.

PR #312, opened 2019-09-11
[node, client] Redis + Webdis lock services

rhlsthrm:
Merge this after it works.


rhlsthrm:
@kthomas implemented distlock algo with Webdis as well.

kthomas:
> @kthomas implemented distlock algo with Webdis as well.

Preference for letting the hub serve as the gatekeeper for the distributed lock.

modules/redis-lock/src/index.ts
```diff
@@ -1,57 +1,55 @@
...
+    // if this function errors out, presumably it is because the lock
+    // could not be acquired. this will bubble up to the caller
+    const lock = await this.redlock.lock(lockName, timeout);
 
-    if (rejectReason) throw new Error(rejectReason);
-
-    return retval;
-  }
+    // run callback
+    const retVal = await callback();
```
kthomas: without the try/catch/finally pattern, if `callback` throws, the lock will not be released until whatever its configured max timeout value is.

kthomas: Disregard this comment, try/catch/finally was added back here: 18a638e4e347133982549c762fc8c555c4deb027

modules/client/src/connext.ts
```diff
@@ -104,6 +114,12 @@ export async function connect(opts: ClientOptions): Promise<ConnextInternal> {
 
   const appRegistry = await node.appRegistry();
 
+  // create the lock service for cfCore
+  // const lockService = new WebdisLockService(webdisUrl);
+  // TODO: switch back to webdis!!
```
bohendo: Goal is to remove webdis entirely, right?

modules/client/src/controllers/ConditionalTransferController.ts
```diff
@@ -207,4 +210,24 @@ export class ConditionalTransferController extends AbstractController {
...
   };
+
+  private async waitForAppInstall(): Promise<void> {
+    return new Promise(
+      async (res: any, rej: any): Promise<any> => {
+        const getAppIds = async (): Promise<string[]> => {
+          return (await this.connext.getAppInstances()).map((a: AppInstanceInfo) => a.identityHash);
+        };
+        let retries = 0;
+        while (!(await getAppIds()).includes(this.appId) && retries <= MAX_RETRIES) {
```
bohendo: These are the kind of pollers that we'll be able to remove once finished events are published right?

modules/client/src/validation/appProposals.ts
```diff
@@ -107,6 +107,7 @@ export const validateLinkedTransferApp = async (
 };
 
 export const appProposalValidation: ProposalValidator = {
+  SimpleLinkedTransferApp: validateLinkedTransferApp,
   SimpleTransferApp: validateSimpleTransferApp,
   SimpleTwoPartySwapApp: validateSwapApp,
   UnidirectionalLinkedTransferApp: validateLinkedTransferApp,
```
bohendo: Apps like `SimpleLinkedTransferApp` are supposed to replace apps like `UnidirectionalLinkedTransferApp` right? So eventually all the unidirectional ones will be removed?

kthomas:
Yes. There should be an IndraLockService instead to proxy lock acquisition.

modules/node/src/redis/redis.provider.ts
```diff
@@ -0,0 +1,51 @@
...
+        console.log("Lost connection to redis. Retrying to connect...");
+        const delay = Math.min(times * 50, 2000);
+        return delay;
+      },
+    });
+    return redisClient;
+  },
+};
+
+export const redlockClientFactory: FactoryProvider = {
```
bohendo: This seems like it should be part of the lock module. Is a separate redis module even needed or is this here just in case we need redis somewhere else for something (YAGNI?)

modules/client/src/controllers/ConditionalTransferController.ts
rhlsthrm: Yep, we will do this before merging (or maybe in a separate PR if we get these tests passing first).

modules/client/src/validation/appProposals.ts
rhlsthrm: Yes, this does replace Unidirectional, we should take that out completely.

modules/node/src/lock/lock.provider.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  async acquireLock(subject: string): Promise<Lock> {
+    const lockName = subject.split(".").pop(); // last item of subject is lock name
+    return await this.lockService.acquireLock(lockName);
+  }
+
+  async releaseLock(subject: string): Promise<Lock> {
+    const lockName = subject.split(".").pop(); // last item of subject is lock name
+    return await this.lockService.acquireLock(lockName);
```
bohendo: Might have meant to say `releaseLock` here

modules/node/src/redis/redis.provider.ts
rhlsthrm: Yes, indeed it is here just so the Redis provider itself is reusable :).

kthomas:
I think it makes decent sense to separate redis client factory from the
lock module.

modules/payment-bot/src/index.ts
```diff
@@ -213,13 +213,15 @@ async function run(): Promise<void> {
...
 
 async function getOrCreateChannel(assetId?: string): Promise<void> {
-  const connextOpts = {
+  const connextOpts: connext.ClientOptions = {
     ethProviderUrl: config.ethProviderUrl,
     logLevel: config.logLevel,
     mnemonic: config.mnemonic,
     nodeUrl: config.nodeUrl,
     store,
+    webdisUrl: "172.17.0.1:6379", // config.webdisUrl,
```
bohendo: We should make this work w config

bohendo: Or removing webdis entirely would be even better

modules/redis-lock/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+    });
+  }
+
+  async acquireLock(
+    lockName: string,
+    callback: (...args: any[]) => any,
+    timeout: number,
+  ): Promise<any> {
+    const lockTTL = 10000;
+    timeout = lockTTL; // HACK-- switch bck to using given timeout
```
bohendo: Why not use given timeout here?

kthomas:
This was a hack. Should use timeout.

modules/redis-lock/src/index.ts
rhlsthrm: Right now we don't know what CF node is passing in, let's confirm our params and then let them pass in their numbers.

modules/client/src/connext.ts
```diff
@@ -55,7 +56,16 @@ import { falsy, notLessThanOrEqualTo, notPositive } from "./validation/bn";
...
 export async function connect(opts: ClientOptions): Promise<ConnextInternal> {
-  const { logLevel, ethProviderUrl, mnemonic, natsClusterId, nodeUrl, natsToken, store } = opts;
+  const {
+    logLevel,
+    ethProviderUrl,
+    mnemonic,
+    natsClusterId,
+    nodeUrl,
+    natsToken,
+    webdisUrl,
```
bohendo: Remind me: This PR is the one that should obsolete anything webdis related right? So this URL should be removed before this PR is merged?

rhlsthrm:
> Yes. There should be an IndraLockService instead to proxy lock acquisition.

https://github.com/ConnextProject/indra/blob/redis-plus-node-lock/modules/redis-lock/src/index.ts#L109

modules/client/src/connext.ts
kthomas: Remove `webdisUrl`. We may need like `indraLockUrl` or similar though..

rhlsthrm: Resolved.

> Remove webdisUrl. We may need like indraLockUrl or similar though..

We don't even, we just use the messaging service that the client already has.

address-book.json
```diff
@@ -224,20 +224,23 @@
...
-    "SimpleTwoPartySwapApp": {
+    "SimpleTransferApp": {
       "address": "0x82D50AD3C1091866E258Fd0f1a7cC9674609D254"
     },
-    "UnidirectionalLinkedTransferApp": {
+    "SimpleTwoPartySwapApp": {
       "address": "0xdDA6327139485221633A1FcD65f4aC932E60A2e1"
     },
-    "UnidirectionalTransferApp": {
+    "UnidirectionalLinkedTransferApp": {
```
bohendo: Are these UnidirectionalApp references on the chopping block too? Should we remove them from the contracts folder too?

modules/client/src/connext.ts
```diff
@@ -55,7 +56,17 @@ import { falsy, notLessThanOrEqualTo, notPositive } from "./validation/bn";
...
 export async function connect(opts: ClientOptions): Promise<ConnextInternal> {
-  const { logLevel, ethProviderUrl, mnemonic, natsClusterId, nodeUrl, natsToken, store } = opts;
+  const {
+    logLevel,
+    ethProviderUrl,
+    mnemonic,
+    natsClusterId,
+    nodeUrl,
+    natsToken,
+    useRedisLock,
```
bohendo: Let's kill this & flag based on presence/absence of the redisUrl instead?

PR #346, opened 2019-09-20
[node, client, contracts] Upgrade cf

LayneHaber:
## The Problem
CF removed some of `ProposeVirtualInstall`

## The Solution
Updated everything to build


modules/types/package.json
```diff
@@ -13,11 +13,11 @@
     "test": "./node_modules/.bin/jest"
   },
   "dependencies": {
+    "@connext/messaging": "0.1.2",
```
ArjunBhuptani: Shouldn't this be updated?

ArjunBhuptani: Also, why do we have messaging in our types?

LayneHaber: ah merge conflict most likely

LayneHaber: fixed

PR #373, opened 2019-09-24
366 withdrawal retry

### LINKED ISSUE #366: [Client, node] Save withdrawal commitment and automatically retry
Right now, if a withdrawal fails (after generating the withdrawal commitment), the user basically has no option except to dispute their channel.

Withdrawal commitment should be saved to store and withdrawal automatically retried twice after 60s to account for cases where the packet to the hub was dropped or lost.

LayneHaber:
Fixes #366


modules/client/src/connext.ts
```diff
@@ -496,6 +504,31 @@ export class ConnextInternal extends ConnextChannel {
...
+    // poll for withdrawal tx submitted to multisig matching tx data
+    const maxBlocks = 15; // wait 15 blocks before retrying/self-submitting
+    const startingBlock = await this.ethProvider.getBlockNumber();
+    let found;
+    interval(async (iteration, stop): Promise<void> => {
+      const currentBlock = await this.ethProvider.getBlockNumber();
+      if (currentBlock <= startingBlock) {
+        return;
+      }
+      found = await this.checkForUserWithdrawal(currentBlock, tx);
```
rhlsthrm: Would we want to use ethers subscriptions/event listening for this instead of polling?

modules/client/src/controllers/WithdrawalController.ts
```diff
@@ -38,7 +38,17 @@ export class WithdrawalController extends AbstractController {
           recipient,
         );
         this.log.info(`Withdraw Response: ${JSON.stringify(withdrawResponse, replaceBN, 2)}`);
-        const minTx = withdrawResponse.transaction;
+        const minTx: CFCoreTypes.MinimalTransaction = withdrawResponse.transaction;
+        // set the withdrawal tx in the store
+        // TODO: better way to set store path
+        await this.connext.store.set([
+          {
+            path: `CF_NODE:store/${this.cfCore.publicIdentifier}/latestNodeSubmittedWithdrawal`,
```
rhlsthrm: In my PR #362 I moved the prefix into a constant in types. We should do that here too. Hopefully my PR gets merged first so you can piggyback.

ArjunBhuptani:
https://github.com/ConnextProject/indra/pull/396 merged, closing

ArjunBhuptani:
Missing tests - can take a crack at it tomorrow

modules/client/src/connext.ts
```diff
@@ -514,6 +553,60 @@ export class ConnextInternal extends ConnextChannel {
...
+  };
+
+  public watchForUserWithdrawal = async (): Promise<void> => {
+    // poll for withdrawal tx submitted to multisig matching tx data
+    const maxBlocks = 15;
+    const startingBlock = await this.ethProvider.getBlockNumber();
+
+    // TODO: poller should not be completely blocking, but safe to leave for now
+    // because the channel should be blocked
+    try {
```
rhlsthrm: I think this try/catch needs to be inside the callback of the subscription, otherwise you are just doing a try/catch on making the subscription itself.

modules/client/src/connext.ts
```diff
@@ -181,6 +188,36 @@ export async function connect(opts: ClientOptions): Promise<ConnextInternal> {
     ...opts, // use any provided opts by default
   });
   await client.registerSubscriptions();
+
+  // make sure there is not an active withdrawal with >= 3 retries
+  const withdrawal = await client.store.get(withdrawalKey(client.publicIdentifier));
+
+  if (withdrawal && withdrawal.retry < 3) {
```
rhlsthrm: Let's move these magic numbers into constants.

modules/client/src/connext.ts
```diff
@@ -1023,6 +1116,37 @@ export class ConnextInternal extends ConnextChannel {
...
+
+  public retryNodeSubmittedWithdrawal = async (): Promise<void> => {
+    const val = await this.getLatestNodeSubmittedWithdrawal();
+    if (!val) {
+      this.logger.error(`No transaction found to retry`);
+      return;
+    }
+    let { retry, tx } = val;
+    retry += 1;
+    if (retry >= 3) {
```
rhlsthrm: Magic numbers to constants.

modules/client/src/connext.ts
```diff
@@ -1023,6 +1116,37 @@ export class ConnextInternal extends ConnextChannel {
...
+      const msg = `Tried to have node submit withdrawal 3 times and it did not work, try submitting from wallet.`;
+      this.logger.error(msg);
+      // TODO: make this submit from wallet :)
+      // but this is weird, could take a while and may have gas issues.
+      // may not be the best way to do this
```
rhlsthrm: Thinking the best way to do this is not to automatically do it, but inform the user and let them do it if they want to spend the gas.

ArjunBhuptani: Not sure how best to do this^ we already log the error to console. It's kind of hard to pass back a return value here

rhlsthrm: Well, we are throwing the error, so we can possibly put the channel in some error state that means it requires user interaction.

modules/client/src/connext.ts
```diff
@@ -599,14 +599,14 @@ export class ConnextInternal extends ConnextChannel {
               `More than ${maxBlocks} have passed, blocks elapsed: ${blockNumber - startingBlock}`,
             );
           }
-        });
+        } catch (e) {
+          if (e.includes(`More than ${maxBlocks} have passed`)) {
```
rhlsthrm: Still confused by this try-catch. What is actually being caught here? Why would the error include the string "More than ${maxBlocks} have passed"? Looking at the `checkForUserWithdrawal` function, I don't see it throwing an error with that message.

modules/client/src/connext.ts
rhlsthrm: Let's put this in a separate issue for now.

modules/client/src/connext.ts
LayneHaber: the promise i create throws that error if more than 15 blocks pass, then pings the hub to resubmit

Issue #336, opened 2019-09-19
ChannelProvider


bohendo:
Push the work started in the channelProvider branch over the finish line

ArjunBhuptani:
We'll try to fit this into the current sprint @pedrouid

hthillman:
Todo: build testing setup (according to @LayneHaber)

LayneHaber:
see PR #375

PR #375, opened 2019-09-26
Channel provider

ArjunBhuptani:
WIP channel provider - still need to set up a way to automagically test this


ArjunBhuptani:
New test strategy: going to just integrate walletconnext into the Dai Card and run two cards

LayneHaber:
@ArjunBhuptani also review this since its been a while since its been updated by you

Will need to update to use latest wc packages from npm once they're published. Since this is thicc want to start the review process now @rhlsthrm @bohendo 

Also not sure the best way to add automated testing for this stuff, but it would be nice to have. Will spend some time looking into this with cypress.

LayneHaber:
Fixes #336

PR #400, opened 2019-10-02
Daicard cleanup

bohendo:
## The Problem

1. Buttons jumped around sometimes
2. Auto-redeeming on link click is dangerous if eg someone is using incognito-mode
3. Need a way for clients to monitor whether link payments have been redeemed yet w/out attempting to redeem them

## The Solution

1. Removed the touch ripple effect & buttons stopped moving
2. New redeem flow involves user clicking through a warning re incognito browsers
3. Added an endpoint + client method for fetching link payment details given a paymentId

Bonus:
 - Setup most components (not App.js though, too complicated) to use react hooks instead of component classes
 - Experimented w `xstate` for managing state of redeem page
 - Refactored the send page to display link payments on it's own modal instead of redirecting to the redeem page w a dont-redeem flag which was kinda weird.
 - Extracted a reusable `Copyable` component to use w all click-to-copy text fields

Potential Backwards Incompatibility: Exposed the channel client's config directly as `channel.config` since it's static & doesn't change. Removed the old `await channel.getConfig()` method.

## Checklist:


modules/client/src/controllers/ResolveConditionController.ts
```diff
@@ -33,23 +35,25 @@ export class ResolveConditionController extends AbstractController {
...
+    this.log.info(`Found link payment for ${amountBN.toString()} ${assetId}`);
 
     const freeBal = await this.connext.getFreeBalance(assetId);
     const preTransferBal = freeBal[this.connext.freeBalanceAddress];
 
     // TODO: dont listen to linked transfer app in default listener, only listen for it here
 
-    await this.node.resolveLinkedTransfer(paymentId, preImage, amount, assetId);
+    const { paymentId, preImage } = params;
+    await this.node.resolveLinkedTransfer(paymentId, preImage);
```
bohendo: The node has info in it's db that links paymentId to the proper amount & assetId so no need for the client to provide either of these derived fields. This also eliminates complexity surrounding case where a user submits mismatched paymentId/amount/assetId

modules/client/src/connext.ts
```diff
@@ -193,10 +196,12 @@ export async function connect(opts: ClientOptions): Promise<ConnextInternal> {
  */
 export abstract class ConnextChannel {
   public opts: InternalClientOptions;
+  public config: GetConfigResponse;
   private internal: ConnextInternal;
 
   public constructor(opts: InternalClientOptions) {
     this.opts = opts;
+    this.config = opts.config;
```
bohendo: Fetch the node's config once on connection and then expose the result to use synchronously later w/out needing to await a msg from the node every time we want to access config. I also removed the async getConfig method in favor of the config property because this config stuff (urls & contract addresses mostly) doesn't change

modules/client/src/node.ts
```diff
@@ -114,15 +110,14 @@ export class NodeApiClient implements INodeApiClient {
...
-    paymentId: string,
-    preImage: string,
-    amount: string,
-    assetId: string,
-  ): Promise<void> {
+  public async fetchLinkedTransfer(paymentId: string): Promise<any> {
+    return await this.send(`transfer.fetch-linked.${this.userPublicIdentifier}`, {
+      paymentId,
+    });
+  }
```
bohendo: New client method added: get link payment details (eg amount, assetId, redemption status) given the paymentId

modules/client/src/types.ts
```diff
@@ -46,15 +47,14 @@ export interface ClientOptions {
...
   contract?: MultisigState;
+  ethProvider: providers.JsonRpcProvider;
   messaging: IMessagingService;
   multisigAddress: string;
   network: utils.Network; // TODO: delete! use bos branch!
   node: NodeApiClient;
-  nodePublicIdentifier: string;
-  ethProvider: providers.JsonRpcProvider;
   store: CFCoreTypes.IStoreService;
-  contractAddresses: ContractAddresses;
```
bohendo: nodePublicIdentifier and contractAddresses are available via config, if we pass the whole config object in then we don't need these separately.

modules/daicard/src/components/AppBar.js
```diff
@@ -28,6 +28,7 @@ export const AppBarComponent = props => (
           >
             <Grid item xs={3}>
               <IconButton
+                disableTouchRipple
```
bohendo: Removing the touch ripple also stops buttons from jumping around

modules/daicard/src/components/Home.js
```diff
@@ -1,137 +1,141 @@
...
-const styles = {};
+const style = withStyles({});
 
-class Home extends React.Component {
-  state = {
-    scanModal: false,
-    history: [],
-  };
+export const Home = style(props => {
+  const [scanModal, setScanModal] = useState(false);
```
rhlsthrm: Yes!

PR #404, opened 2019-10-02
Buidler migration demo

alcuadrado:
Hey Connext team! Here's a silly breaking test case on a Buidler setup so you can give Buidler a try.

You can learn more about the migration changes I made in this Waffle on Buidler guide https://buidler.dev/guides/waffle-testing.html.

Enjoy!


rhlsthrm:
We'll get this implemented when we get back to Solidity dev!

PR #406, opened 2019-10-03
Authentication

bohendo:
## The Problem

Anyone could request the hub to take any action on any channel.. and the hub would just do it, no questions asked.

## The Solution

Verify signatures to ensure that any user who asks to take an action on a given channel, controls the keys for the xpub associated w that channel.

Locks are designated by multisig address so the new auth mechanism ensures the sig is for the address associated with the xpub associated with the multisig address.

Potential Backwards Incompatibilities:
 - I'm no longer using `@connext/proxy-lock`, instead I added the exact same logic to the client's NodeApiClient so that the lock subject messaging can share auth logic w that protecting channel & transfer subjects.

## Checklist:


modules/client/src/node.ts
```diff
@@ -19,6 +19,9 @@ import { Logger } from "./lib/logger";
 import { replaceBN } from "./lib/utils";
 import { NodeInitializationParameters } from "./types";
 
+// Include our access token when interacting with these subjects
+const guardedSubjects = ["channel", "transfer"];
```
kthomas: The subjects you care about have to be supplied when you connect to NATS by supplying the signed JWT in the CONNECT request.

```
[94950] 2019/10/03 03:29:53.022012 [INF] Starting nats-server version 2.1.0
[94950] 2019/10/03 03:29:53.022096 [DBG] Go build version go1.12.6
[94950] 2019/10/03 03:29:53.022101 [INF] Git commit [not set]
[94950] 2019/10/03 03:29:53.022357 [INF] Listening for client connections on 0.0.0.0:4220
[94950] 2019/10/03 03:29:53.022365 [INF] Server id is NBSKPK2DWYSQSPSQ3X5QQ4VQXXLJ6EVA272H6HHIOOGRUBLGDECD6PS4
[94950] 2019/10/03 03:29:53.022367 [INF] Server is ready
[94950] 2019/10/03 03:29:53.022376 [DBG] Get non local IPs for "0.0.0.0"
[94950] 2019/10/03 03:29:53.022553 [DBG]  ip=192.168.254.79
[94950] 2019/10/03 03:30:13.338485 [DBG] 127.0.0.1:59432 - cid:1 - Client connection created
[94950] 2019/10/03 03:30:13.355022 [TRC] 127.0.0.1:59432 - cid:1 - <<- [CONNECT {"lang":"typescript","version":"1.2.4","verbose":false,"pedantic":false,"protocol":1,"name":"ts-natsutil-4bfd6be6-1377-470a-b462-23738541b7f7","jwt":"<INSERT SIGNED JWT WITH PERMISSIONS>"}]
```

modules/node/src/auth/auth.service.ts
```diff
@@ -0,0 +1,99 @@
...
+  return { err: `Invalid token` } as any;
+};
+
+const badSubject = (warning: string): any => {
+  logger.warn(warning);
+  return { err: `Invalid subject` } as any;
+};
+
+@Injectable()
+export class AuthService {
```
kthomas: This is probably a good place to add the JWT vending.

The JWT needs to include the following claims:
```
{
  "iat":1570112805,
  "exp":1601648805,
  "aud":"nats.connext.network",
  "sub":"someXpub",
  "permissions":[

  ]
}
```

Insert NATS `permissions` for publish, subscribe and request/response. Then sign it.

kthomas: `aud` isn't really needed above.

PR #437, opened 2019-10-16
daicard migration

### LINKED ISSUE #425: [card] Clicking p2p-send if recipient is left null will transfer to 0x00000.. :(
**Describe the bug**
Logs after I think I just burnt some rinkeby $$$:
```
info: [ConditionalTransferController] Conditional transfer called with parameters: {
  "assetId": "0xFab46E002BbF0b4509813474841E0716E6730136",
  "amount": "10000000000000000",
  "conditionType": "LINKED_TRANSFER_TO_RECIPIENT",
  "paymentId": "0xa57b9b8c8ceed6d6d8e2c4aad1926fcb05516294b04776e56ad38fb3719dacab",
  "preImage": "0xd455a011ebb27f008b2167321e4063cbbbbc90d1e819eb6097f6f31046e22f3d",
  "recipient": null
}
```

We should guard against this at the UI level (obviously) but probably also at the client level too..

### LINKED ISSUE #431: [card] Floor balance instead of rounding
**Describe the bug**
It's confusing if the user's balance is 1.495, it'll display as "1.50" but the user won't actually be able to send "1.50" dai. We should floor instead of rounding the display to avoid these cases.

### LINKED ISSUE #435: [card] Build "upgrade" path in v2 dai card for existing v1 users
We chatted about this in person at Devcon - Bo's suggestion was to have an automated UI prompt for any existing Dai Card (v1) which forces the user to "upgrade".

When the user hits the upgrade button, funds are withdrawn from the v1 Dai Card and then deposited to the v2 one.

Things we'll need for this:
- Localstorage variable to keep track of whether to show the v1 upgrade prompt or not
- UI component which the user clicks to upgrade (plus text such as "withdrawing from v1", "depositing to v2" so that users know what they're waiting on)
- Integrating v1 client into the v2 Dai Card and wiring up withdrawal

bohendo:
## The Problem

closes #435
closes #425 
closes #431 

## Checklist:


modules/daicard/src/utils/migrate.js
```diff
@@ -0,0 +1,51 @@
...
+    console.log(`Cashing out legacy channel: ${JSON.stringify(
+      withdrawalParams,
+      (key, value) => value._hex ? toBN(value._hex).toString() : value,
+      2,
+    )}`);
+
+    try {
+      // To debug: simulate a withdrawal by just waiting for a bit
+      await new Promise((res, rej) => setTimeout(res, 5000));
+      // await legacy.withdraw(withdrawalParams);
```
LayneHaber: im not clear on what this is simulating or how it empties the channel? seems like this would have to be done for the migrations

LayneHaber: what happens if someone refreshes *during* a migration, would it just freeze up and say that it's already migrating but never migrate?

PR #439, opened 2019-10-18
Daicard proxy

bohendo:
## The Problem

Indra lives at indra.connext.network and this is the endpoint we want integrations to use when interacting with our hub.

The daicard app (and all our user mnemonics) live at daicard.io.

Can't have one server serve two domains.

## The Solution

Create a lightweight proxy at daicard.io that does as little as possible besides forwarding requests from daicard.io to indra.connext.network.

New `modules/proxy/daicard.io/entry.sh` is basically a copy/paste of the old one (that's now in `modules/proxy/indra.connext.network`). It's mostly responsible for setting up https certs but also waits for upstream services.

New `modules/proxy/daicard.io/prod.conf` is very similar to indra's. The root rule (`location / { ... }`) just forwards everything to the upstream indra url. The `/api/messaging` rule special bc it uses ws to talk to nats so can't just blindly proxy pass it. Gotta speak the ws language while passing traffic to Indra's ws-nats relay.

## Checklist:


bohendo:
New `modules/proxy/daicard.io/entry.sh` is basically a copy/paste of the old one (that's now in `modules/proxy/indra.connext.network`). It's mostly responsible for setting up https certs but also waits for upstream services.

New `modules/proxy/daicard.io/prod.conf` is very similar to indra's. The root rule (`location / { ... }`) just forwards everything to the upstream indra url. The `/api/messaging` rule special bc it uses ws to talk to nats so can't just blindly proxy pass it. Gotta speak the ws language while passing traffic to Indra's ws-nats relay.

modules/proxy/daicard.io/entry.sh
```diff
@@ -0,0 +1,84 @@
...
+fi
+
+echo "Using certs for $domain"
+ln -sf $letsencrypt/$domain/privkey.pem /etc/certs/privkey.pem
+ln -sf $letsencrypt/$domain/fullchain.pem /etc/certs/fullchain.pem
+
+# Hack way to implement variables in the nginx.conf file
+sed -i 's/$hostname/'"$domain"'/' /etc/nginx/nginx.conf
+sed -i 's|$INDRA_URL|'"$indra_url"'|' /etc/nginx/nginx.conf
+sed -i 's|$RELAY_URL|'"http://${indra_url#*://}:4223"'|' /etc/nginx/nginx.conf
```
bohendo: This doesn't work w https, not sure why.. idk how big of a problem it is to use http here..

PR #442, opened 2019-10-18
Daicard state unification

bohendo:
NOT READY TO MERGE YET: lost the txHash in the notification but other than that it's ready to be reviewed.

Inspiration for this PR brought to you by the EthBerlin hackathon. Where I spent all weekend re-integrating the daicard logic into an app w a very different layout. Which was annoying bc a lot of the daicard logic is really generic and will likely be used by many apps that integrate our platform (eg the TipDai bot..) Separating our app's core logic from the react components would go a long way in making the daicard something that can be easily plugged into other apps.

This PR is step 1 of a longer and, honestly, relatively low priority process. It features a refactor of the app's root state to use one xstate machine instead of handful of random react state vars.

Using xstate gives us access to some cool tools, eg if you copy/paste the [root xstate machine](https://github.com/ConnextProject/indra/blob/f74f83583c08fe91786f6ce0a69f28a769182f5d/modules/daicard/src/state/root.js) into the [visualizer](https://xstate.js.org/viz/), you can generate some nify state graphs. Eg:

Screenshot from 2019-10-18 21-54-42


rhlsthrm:
This is awesome! Definitely looking forward to seeing this come together!

ArjunBhuptani:
Really really cool and will definitely be really helpful as the card gets used as a reference implementation more often :D

PR #466, opened 2019-10-22
[node, client] Integrate app CF app-specific keys - DO NOT MERGE

### LINKED ISSUE #463: [client, protocol, node, card] Improve mnemonic security
There's two interconnected ways that we can improve the security of mnemonics in our system.

First, @bohendo's suggestion was to stop storing mnemonic in the store/db and have it be only used in-memory. This would involve a deep refactor to change how mnemonics are retrieved in the CF module in addition to a breaking change to how the CF module is instantiated (passing in mnemonic instead of store with mnemonic in it).

Second, we've discussed for some time that we need to add support for hosting the mnemonic OUTSIDE the Connext client and passing in a keygen fn which can spit out app-specific ephemeral keys as needed. This would involve a refactor of how keys are generated/retrieved in the CF module in addition to a breaking change to how the CF module is instantiated (to separate out passing the mnemonic exactly like above and to optionally support the keygen fn instead of mnemonic).

Note that the latter is needed for wallet integrations with Metamask, Mosendo, Torus, Portis, etc.  since none of them want to expose mnemonics to Connext code.

There's obviously a lot of overlap here so I think it makes sense to do this as one PR to the CF code altogether + a corresponding singular PR to ours. What do you guys think?

rhlsthrm:
## The Problem
No longer need to pass mnemonic into CF core. Fixes #463 

## The Solution
Integrated new CF version

## Checklist:


rhlsthrm:
Do not merge until #493 is fixed in this branch, otherwise all live channels will break.

modules/client/src/connext.ts
```diff
@@ -83,9 +78,22 @@ const MAX_WITHDRAWAL_RETRIES = 3;
...
+  let keyGen;
+  if (mnemonic) {
+    const hdNode = fromExtendedKey(fromMnemonic(mnemonic).extendedKey).derivePath(CF_PATH);
+    xpub = hdNode.neuter().extendedKey;
+    keyGen = (index: string): Promise<string> =>
+      Promise.resolve(hdNode.derivePath(index).privateKey);
+  } else {
+    xpub = opts.xpub;
+    keyGen = opts.keyGen;
+  }
```
LayneHaber: we should throw errors if these aren't defined, typescript won't catch it

modules/client/src/connext.ts
```diff
@@ -626,7 +622,7 @@ export class ConnextInternal extends ConnextChannel {
     // TODO: poller should not be completely blocking, but safe to leave for now
     // because the channel should be blocked
     try {
-      await new Promise((resolve, reject) => {
+      await new Promise((resolve: any, reject: any): any => {
```
LayneHaber: can make this return void, should also add a type that is PromiseRes and PromiseRej since we do this a *lot*

modules/client/src/connext.ts
```diff
@@ -671,45 +667,58 @@ export class ConnextInternal extends ConnextChannel {
...
 
     // try to recover the rest of the stateS
-    try {
-      await this.restoreStateFromBackup(xpub);
-      this.logger.debug(`restored state from backup!`);
-    } catch (e) {
-      await this.restoreStateFromNode(xpub);
-      this.logger.debug(`restored state from node!`);
-    }
+    // TODO: fix pisa implementation
```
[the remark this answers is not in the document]
LayneHaber: (since virtual channels will likely depend on a pisa restoration)

modules/client/src/connext.ts
```diff
@@ -1245,7 +1264,8 @@ export class ConnextInternal extends ConnextChannel {
       this.logger.error(`No transaction found to retry`);
       return;
     }
-    let { retry, tx } = val;
+    let { retry } = val;
+    const { tx } = val;
```
LayneHaber: all about them ts-lint fixes this PR, respek

modules/client/src/controllers/ConditionalTransferController.ts
```diff
@@ -65,7 +65,7 @@ export class ConditionalTransferController extends AbstractController {
     const ret = await this.handleLinkedTransfers(params);
 
     // set recipient and encrypted pre-image on linked transfer
-    const recipientPublicKey = fromExtendedKey(recipient).publicKey;
+    const recipientPublicKey = fromExtendedKey(recipient).derivePath("0").publicKey;
```
LayneHaber: can we replace this with `xpubToAddress` fn (and any other time we do this)

modules/node/src/cfCore/cfCore.provider.ts
```diff
@@ -2,19 +2,29 @@ import { IMessagingService, MessagingServiceFactory } from "@connext/messaging";
...
 import { CLogger, freeBalanceAddressFromXpub } from "../util";
-import { CFCore, EXTENDED_PRIVATE_KEY_PATH } from "../util/cfCore";
+import { CFCore } from "../util/cfCore";
 
 import { CFCoreRecordRepository } from "./cfCore.repository";
 
 const logger = new CLogger("CFCoreProvider");
 
+// TODO: where should this live?
+const LATEST_CF_STORE_VERSION = 1;
```
LayneHaber: our latest version should live in a table, the latest version they support should live in the node. then we should make sure we note if migrations have been performed in our table as well

modules/node/src/channel/channel.service.ts
```diff
@@ -239,6 +239,40 @@ export class ChannelService {
     return txRes;
   }
 
+  async migrateLegacyChannel(userPublicIdentifier: string): Promise<void> {
```
LayneHaber: how will this fn be called?

will probably want to make this more general, since the "legacy channel" format will depend on store versions we are maintaining vs. store versions the client has (im assuming these store versions are client driven atm)

modules/node/src/constants.ts
```diff
@@ -35,3 +35,6 @@ export const defaultPaymentProfileEth: PaymentProfile = {
   id: 0,
   minimumMaintainedCollateral: parseEther("0.05"),
 };
+
+// CF
+export const CF_PATH = "m/44'/60'/0'/25446";
```
LayneHaber: tbh we should just put this in the types package

rhlsthrm:
@LayneHaber, I got the store entries separated out when storing to Postgres. I haven't implemented this on the client-side yet, which may not be as important (only one channel on the client). This should be ready to test with migrations.

rhlsthrm:
I did this in JS now because I couldn't open Discord to see @hthillman's SQL queries he sent me. We should definitely refactor this to SQL. Also, this makes me realize we can already add a nice schema into our DB and simply deconstruct/reconstruct it in the get/set function. I'll work on the SQL bits tomorrow, but it shouldn't change the interface. I'll collab with @hthillman on the best way to structure the schema.

modules/client/src/connext.ts
```diff
@@ -83,9 +78,22 @@ const MAX_WITHDRAWAL_RETRIES = 3;
...
 
 export async function connect(opts: ClientOptions): Promise<ConnextInternal> {
-  const { logLevel, ethProviderUrl, mnemonic, natsClusterId, nodeUrl, natsToken, store } = opts;
+  const { ethProviderUrl, logLevel, mnemonic, natsClusterId, natsToken, nodeUrl, store } = opts;
   const logger = new Logger("ConnextConnect", logLevel);
 
+  // Convert mnemonic into xpub + keyGen if provided
+  let xpub: string;
+  let keyGen;
+  if (mnemonic) {
```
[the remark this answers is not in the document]
rhlsthrm: Yes, we decided to keep it in more for the quick testing aspect.

modules/client/src/connext.ts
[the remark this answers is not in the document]
rhlsthrm: Just fixing the store to send the correct entries to Pisa. It's not too hard, but it just wasn't a priority at the moment. I think it's higher priority now that restore is messed up in some cases.

modules/client/src/listener.ts
```diff
@@ -334,8 +308,26 @@ export class ConnextListener extends EventEmitter {
     const subject = `transfer.send-async.${this.connext.publicIdentifier}`;
     await this.connext.messaging.subscribe(subject, async (data: any) => {
       this.log.info(`Received message for subscription: ${JSON.stringify(data)}`);
-      const { encryptedPreImage, paymentId } = data;
+      let paymentId: string;
+      let encryptedPreImage: string;
+      if (data.paymentId) {
```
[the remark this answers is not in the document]
rhlsthrm: Yes, looks like it varies based on websockets vs not.

modules/daicard/src/App.js
```diff
@@ -46,6 +47,7 @@ const urls = {
 const WITHDRAW_ESTIMATED_GAS = toBN("300000");
 const DEPOSIT_ESTIMATED_GAS = toBN("25000");
 const MAX_CHANNEL_VALUE = Currency.DAI("30");
+const CF_PATH = "m/44'/60'/0'/25446";
```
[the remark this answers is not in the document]
rhlsthrm: Yes we should. I think I won't do that in this PR though.

modules/node/src/cfCore/cfCore.provider.ts
rhlsthrm: Why does our latest version need to live in a table? We will have to change app code every time we want to migrate so it seems like it should be in the code itself. Also, they are building a more robust migration framework so we should probably wait for now.

modules/node/src/channel/channel.service.ts
rhlsthrm: This was a method I started working on before Liam made the migration script, so this isn't needed anymore.

modules/client/src/controllers/ConditionalTransferController.ts
rhlsthrm: Tbh, `xpubToAddress` is a kind of confusing name. Maybe it should be `xpubToZeroAddress` or something.

rhlsthrm: Realizing we can't actually replace this because we need the public key not the ETH address.

LayneHaber:
@rhlsthrm should we close this PR out?

We'll probably want the store stuff from this branch relatively soon. I at least heavily prefer the restructuring of the data, but maybe not the in memory stuff

rhlsthrm:
Yes, we'll close this branch out, it's obviated now.

> I at least heavily prefer the restructuring of the data, but maybe not the in-memory stuff

Are you talking about the CF changes? Let's pick out exactly which ones we want.

PR #524, opened 2019-11-05
502 restore state earlier

bohendo:
## The Problem

On startup, the client tries to redeem link payments, etc.

If the state needs to be restores, that needed to happen after calling connect.

But the async redemption of connect failed until state could be restored.

Not a great situation..

## The Solution

State is now being restored on connect if necessary.

This required adding a restart method to the client to restart the cfCore piece after the state has been updated.

While I was at it, I moved the client interface into types and did some serious types cleanup to move all definitions out of the big index & into the subunits.

Backwards incompatible: Added a new function to the client interface and changed the signature of restoreState.

## Checklist:


modules/client/src/controllers/TransferController.ts
```diff
@@ -180,12 +177,12 @@ export class TransferController extends AbstractController {
         boundResolve = this.resolveInstallTransfer.bind(null, res);
         this.listener.on(CFCoreTypes.EventName.INSTALL_VIRTUAL, boundResolve);
         this.listener.on(CFCoreTypes.EventName.REJECT_INSTALL_VIRTUAL, boundReject);
-        // this.timeout = setTimeout(() => {
-        //   this.cleanupInstallListeners(boundResolve, boundReject);
-        //   boundReject({ data: { appInstanceId: this.appId } });
-        // }, 5000);
+        this.timeout = setTimeout((): void => {
+          this.cleanupInstallListeners(boundResolve, boundReject);
+          boundReject({ data: { appInstanceId: this.appId } });
```
LayneHaber: wont reject properly:
```
// check app id
if (this.appId !== (msg as any).appInstanceId) {	    
      return;
}
```

modules/client/src/controllers/WithdrawalController.ts
```diff
@@ -47,12 +47,14 @@ export class WithdrawalController extends AbstractController {
...
-        this.log.info(`Node Withdraw Response: ${JSON.stringify(transaction, replaceBN, 2)}`);
+        this.log.info(`Node Withdraw Response: ${stringify(transaction)}`);
       } else {
         this.log.info(`Calling ${CFCoreTypes.RpcMethodName.WITHDRAW}`);
-        const withdrawResponse = await this.connext.cfWithdraw(amount, assetId, recipient);
-        this.log.info(`Withdraw Response: ${JSON.stringify(withdrawResponse, replaceBN, 2)}`);
-        transaction = await this.ethProvider.getTransaction(withdrawResponse.txHash);
+        // Bo note: the following function used to point to something that doesn't exist
+        // so this code path probably never gets executed.. idk what this was supposed to do
+        const withdrawResponse = await this.node.withdraw(transaction as any);
```
LayneHaber: this was to allow the user to submit their own withdrawal, without it going through the hub to pay gas. we should probably keep it in here :)

modules/client/src/types.ts
```diff
@@ -2,75 +2,38 @@ import { IMessagingService } from "@connext/messaging";
...
-}
+export type BigNumber = connextBN;
+export const BigNumber = connextBN;
 
 export type InternalClientOptions = ClientOptions & {
   appRegistry: AppRegistry;
   channelRouter: ChannelRouter;
+  channelProvider?: ChannelProvider;
   config: GetConfigResponse;
   contract?: MultisigState;
```
LayneHaber: hm, i know this isnt with this PR but not sure how i didn't see this earlier, what is this:
```typescript
type MultisigState<T = string> = {
    id: number;
    xpubA: string;
    xpubB: string;
    multisigAddress: string;
    freeBalanceA: T;
    freeBalanceB: T;
    appIds: number[];
}
```

the free balance would never be just a `string` or a `BigNumber` and im not sure how it's being used here or in instantiation?

also, why is the appId a `number[]`? the only IDs we ever use are strings

Is this something that gets pulled by the hub? can we delete?

modules/client/src/validation/appProposals.ts
```diff
@@ -186,26 +176,26 @@ const baseAppValidation = async (
     initiatorFreeBalance[freeBalanceAddressFromXpub(connext.nodePublicIdentifier)];
   if (isVirtual && nodeFreeBalance.lt(app.initiatorDeposit)) {
     const reqRes = await connext.requestCollateral(app.initiatorDepositTokenAddress);
-    connext.logger.info(`Collateral Request result: ${JSON.stringify(reqRes, replaceBN, 2)}`);
+    connext.log.debug(`Collateral Request result: ${reqRes}`);
```
LayneHaber: `result: ${stringify(reqRes)}`

modules/payment-bot/src/index.ts
```diff
@@ -218,16 +167,22 @@ async function run(): Promise<void> {
...
+        );
+        console.log(`Found postwithdrawal balance of ${formatEther(postWithdrawBal)}`);
+      },
+    );
+    client.on(
+      CFCoreTypes.EventName.WITHDRAWAL_FAILED,
+      async (data: any): Promise<void> => {
+        console.log(`Withdrawal failed with data: ${JSON.stringify(data, replaceBN, 2)}`);
+      },
+    );
```
LayneHaber: we should hard error if any of these `failed` events come through, since as it stands bots will continue to execute and CI won't catch it

modules/payment-bot/src/store.ts
```diff
@@ -97,83 +97,18 @@ export class Store implements CFCoreTypes.IStoreService {
...
-      // see: https://github.com/counterfactual/monorepo/issues/2497
-      if (doh.message) {
-        const matches = doh.message.match(Store.PisaNonceTooLowRegex);
-        if (matches && Number.parseInt(matches[1], 10) === version) {
-          console.error(doh);
-          console.error(stringed);
-          return;
-        }
-      }
-      throw doh;
```
LayneHaber: sad to see this go tbh

modules/types/src/basic.ts
```diff
@@ -0,0 +1,21 @@
...
+export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+
+export type SolidityValueType = any; // FIXME: use cf type
+
+export const ConnextEvents = {
+  RECIEVE_TRANSFER_FAILED: "receiveTransferFailedEvent",
+  RECIEVE_TRANSFER_FINISHED: "receiveTransferFinishedEvent",
+  RECIEVE_TRANSFER_STARTED: "receiveTransferStartedEvent",
+};
+export type ConnextEvent = keyof typeof ConnextEvents;
```
LayneHaber: we really need to extend the cf events as well (since the client emits those additionally)

LayneHaber: obviously not blocking the merge though, ill make an issue

modules/types/src/channel.ts
```diff
@@ -0,0 +1,87 @@
...
+import { App, AppState } from "./app";
+import { Address } from "./basic";
+
+////////////////////////////////////
+////// CHANNEL TYPES
+
+// used to verify channel is in sequence
+export type ChannelAppSequences = {
+  userAppSequenceNumber: number;
+  nodeAppSequenceNumber: number;
```
LayneHaber: we should just call these `userSequenceNumber` rather than `userAppSequenceNumber`

modules/types/src/channel.ts
```diff
@@ -0,0 +1,87 @@
...
+export interface AssetAmount<T = string> {
+  amount: T;
+  assetId: Address; // empty address if eth
+}
+export type AssetAmountBigNumber = AssetAmount<BigNumber>;
+
+export type User = {
+  id: number;
+  xpub: string;
+  channels: CFCoreChannel[];
```
LayneHaber: when is this type used? when would a user have more than a single channel?

modules/types/src/channel.ts
```diff
@@ -0,0 +1,87 @@
...
+export type MultisigState<T = string> = {
+  id: number;
+  xpubA: string;
+  xpubB: string;
+  multisigAddress: string;
+  freeBalanceA: T;
+  freeBalanceB: T;
+  appIds: number[];
+};
+export type MultisigStateBigNumber = MultisigState<BigNumber>;
```
LayneHaber: reposting:
hm, i know this isnt with this PR but not sure how i didn't see this earlier, what is this:

```typescript
type MultisigState<T = string> = {
    id: number;
    xpubA: string;
    xpubB: string;
    multisigAddress: string;
    freeBalanceA: T;
    freeBalanceB: T;
    appIds: number[];
}
```
the free balance would never be just a string or a BigNumber and im not sure how it's being used here or in instantiation?

also, why is the appId a number[]? the only IDs we ever use are strings

Is this something that gets pulled by the hub? can we delete?

modules/types/src/inputs.ts
```diff
@@ -72,17 +72,19 @@ export type ResolveConditionResponse<T = string> = ResolveLinkedTransferResponse
...
 export type TransferCondition = keyof typeof TransferConditions;
 
 // linked transfer types
 export type LinkedTransferParameters<T = string> = {
-  conditionType: "LINKED_TRANSFER";
-  amount: T;
-  assetId: Address; // make optional?
+  conditionType: TransferCondition;
+  amount?: T;
+  assetId?: Address;
```
LayneHaber: when are these optional?

modules/types/src/basic.ts
LayneHaber: https://github.com/ConnextProject/indra/issues/525

modules/types/src/inputs.ts
LayneHaber: assetId can default to ETH, but there should always be a defined amount

modules/client/src/controllers/TransferController.ts
rhlsthrm: We shouldn't do it this way, using Promise.race seems a lot cleaner:

https://github.com/ConnextProject/indra/pull/466/files?file-filters%5B%5D=.ts#diff-d47b6d123be3db39095d99eea5357c64R178

modules/client/src/types.ts
LayneHaber: i deleted it

modules/types/src/channel.ts
LayneHaber: i deleted it

modules/client/src/controllers/WithdrawalController.ts
```diff
@@ -50,11 +50,14 @@ export class WithdrawalController extends AbstractController {
         this.log.info(`Node Withdraw Response: ${stringify(transaction)}`);
       } else {
         this.log.info(`Calling ${CFCoreTypes.RpcMethodName.WITHDRAW}`);
-        // Bo note: the following function used to point to something that doesn't exist
-        // so this code path probably never gets executed.. idk what this was supposed to do
-        const withdrawResponse = await this.node.withdraw(transaction as any);
```
bohendo: Yay thank you @LayneHaber for replacing this with what it was supposed to be!

PR #546, opened 2019-11-12
Walletconnect local storage

LayneHaber:
makes sure wc works properly on reloads. dont merge until wc is updated


modules/client/src/connext.ts
```diff
@@ -900,10 +900,11 @@ export class ConnextClient implements IConnextClient {
   };
 
   public matchTx = (
-    givenTransaction: Transaction,
+    givenTransaction: Transaction | undefined,
     expected: CFCoreTypes.MinimalTransaction,
   ): boolean => {
     return (
+      givenTransaction &&
       givenTransaction.to === expected.to &&
```
ArjunBhuptani: Will this work if `givenTransaction` is `undefined`?

LayneHaber: yessir

PR #548, opened 2019-11-13
[node] Collateral checks

LayneHaber:
## The Problem
There wasn't a system to ensure the hub has sufficient collateral in a channel before resolving a linked transfer.

## The Solution
If the hub needs to deposit, it will create a promise thats resolved on its next deposit event into the users multisig. Otherwise, it will request collateral as usual without `await` -ing.

## Checklist:

### DIFF
--- a/modules/cf-core/src/methods/app-instance/propose-install/controller.ts
+++ b/modules/cf-core/src/methods/app-instance/propose-install/controller.ts
@@ -23,11 +23,14 @@ import {
 export default class ProposeInstallController extends NodeController {
   @jsonRpcMethod(Node.RpcMethodName.PROPOSE_INSTALL)
   @jsonRpcMethod(Node.RpcMethodName.PROPOSE_INSTALL_VIRTUAL)
-  public executeMethod = super.executeMethod;
+  public executeMethod: (
+    requestHandler: RequestHandler,
+    params: Node.MethodParams,
+  ) => Promise<Node.MethodResult> = super.executeMethod;
 
   protected async getRequiredLockNames(
     requestHandler: RequestHandler,
-    params: Node.ProposeInstallParams
+    params: Node.ProposeInstallParams,
   ): Promise<string[]> {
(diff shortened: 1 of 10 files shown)


modules/cf-core/src/methods/app-instance/propose-install/controller.ts
```diff
@@ -23,28 +23,31 @@ import {
 export default class ProposeInstallController extends NodeController {
   @jsonRpcMethod(Node.RpcMethodName.PROPOSE_INSTALL)
   @jsonRpcMethod(Node.RpcMethodName.PROPOSE_INSTALL_VIRTUAL)
-  public executeMethod = super.executeMethod;
+  public executeMethod: (
```
rhlsthrm: Let's revert all the whitespace changes.

modules/node/src/transfer/transfer.service.ts
```diff
@@ -201,16 +202,47 @@ export class TransferService {
       throw new Error(`App with provided hash has not been installed: ${linkedHash}`);
     }
 
+    const freeBalanceAddr = this.cfCoreService.cfCore.freeBalanceAddress;
+
     const freeBal = await this.cfCoreService.getFreeBalance(
```
rhlsthrm: Just realized we should make helper functions at the channel service level to wrap `getNodeFreeBalance`.

modules/node/src/transfer/transfer.service.ts
```diff
@@ -201,16 +202,47 @@ export class TransferService {
...
+            }
+            if (msg.data.multisigAddress !== channel.multisigAddress) {
+              return;
+            }
+            resolve();
+          },
+        );
+        try {
+          const result = await this.channelService.requestCollateral(userPubId, assetId, amountBN);
+          if (!result) {
```
rhlsthrm: Maybe we should check something like `if (typeof result !== Promise && !result) ...` just for safety?

PR #550, opened 2019-11-13
[node, cf-core, cf-types, client] Needed some fixes for the multisig issues, added the propose protocol

rhlsthrm:
## Checklist:

### DIFF
--- a/modules/cf-core/src/protocol/install-virtual-app.ts
+++ b/modules/cf-core/src/protocol/install-virtual-app.ts
@@ -20,10 +20,10 @@ import {
   TwoPartyFixedOutcomeInterpreterParams,
   virtualAppAgreementEncoding,
 } from "../types";
-import { getCreate2MultisigAddress } from "../utils";
 
 import { UNASSIGNED_SEQ_NO } from "./utils/signature-forwarder";
 import { assertIsValidSignature } from "./utils/signature-validator";
+import { Store } from "../store";
 
 export const encodeSingleAssetTwoPartyIntermediaryAgreementParams = params =>
   defaultAbiCoder.encode([virtualAppAgreementEncoding], [params]);
@@ -56,7 +56,8 @@ export const INSTALL_VIRTUAL_APP_PROTOCOL: ProtocolExecutionFlow = {
       message: { params, processID },
       stateChannelsMap,
       network,
-      provider
+      provider,
+      store
     } = context;
 
     const {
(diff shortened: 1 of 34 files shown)


modules/cf-core/src/store.ts
```diff
@@ -215,7 +216,8 @@ export class Store {
     myIdentifier: string,
     theirIdentifier: string,
     proxyFactoryAddress: string,
-    minimumViableMultisigAddress: string
+    minimumViableMultisigAddress: string,
+    acceptGeneratedMultisig: boolean // = false,
```
rhlsthrm: Why not default to false?

modules/cf-core/src/protocol/install-virtual-app.ts
```diff
@@ -320,7 +322,8 @@ export const INSTALL_VIRTUAL_APP_PROTOCOL: ProtocolExecutionFlow = {
       stateChannelsMap,
       (virtualAppInstanceIdentityHash as unknown) as string,
       (virtualAppInstanceDefaultOutcome as unknown) as string,
-      network
+      network,
+      store,
```
rhlsthrm: Do we need this anymore?

modules/cf-core/src/store.ts
LayneHaber: that was temporary, added so the compiler would  point me to all uses. I wanted to go through them one by one in this case, but it now does default to false

modules/cf-core/src/protocol/install-virtual-app.ts
LayneHaber: nope, removed and using static fn

LayneHaber:
alrighty so to squash the multisig address/proxy factory issues and prevent any issues coming up with bytecode things ive done the following:
- removed any calls to getCreate2MultisigAddress accept for within the store class
- added a static method to the store class so we can access this functionality without including the store in the protocol context
- add a boolean flag to the getMultisigAddressWithCounterparty to the store, if true it will accept the address returned from the getCreate2... fn, and if false, it will only take the values from the store
- set the flag to default false, EXCEPT in the following cases:
1. create
2. propose install (no clear indicator of whether it is a virtual or regular proposal, and since most initiator/responder multisigs will not exist we cant enforce it come from the store)
3. un- and install virtual for multisigs between all three participants, or initiator/responder (not intermediary) channels

Issue #467, opened 2019-10-22
Node crashes if attempting to create link w negative amount


bohendo:
```
(node:89) UnhandledPromiseRejectionWarning: Error: invalid BigNumber value (arg="value", value={"_hex":"-0xb1a2bc2ec50000"}, version=4.0.37)
    at Object.throwError (/root/node_modules/ethers/errors.js:76:17)
    at new BigNumber (/root/node_modules/ethers/utils/bignumber.js:99:20)
    at Object.bigNumberify (/root/node_modules/ethers/utils/bignumber.js:182:12)
    at new bigNumberify (/root/node_modules/@counterfactual/node/src/models/app-instance-proposal.ts:80:29)
    at handleReceivedProposalMessage (/root/node_modules/@counterfactual/node/src/message-handling/handle-node-message.ts:29:20)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:89) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
```

We should catch this edge case early & handle it more gracefully

PR #555, opened 2019-11-14
[client, types] Fix negative big number error, update validation library

rhlsthrm:
## The Problem
#467

## The Solution
Added some fixes around the validation library to reduce boilerplate.

## Checklist:


modules/client/src/controllers/TransferController.ts
```diff
@@ -28,20 +27,20 @@ export class TransferController extends AbstractController {
...
-    if (invalid) {
-      throw new Error(invalid.toString());
-    }
+    const freeBalance = await this.connext.getFreeBalance(assetId);
+    const preTransferBal = freeBalance[this.connext.freeBalanceAddress];
+    const errs = [
+      invalidXpub(recipient),
+      invalidAddress(assetId),
+      notLessThanOrEqualTo(amount, preTransferBal),
+    ];
```
LayneHaber: shouldnt this be using `validate`?

modules/types/src/convert.ts
```diff
@@ -35,9 +35,19 @@ const getType = (input: any): NumericTypeName => {
 
 const castFunctions: any = {
   "bignumber-str": (x: BigNumber): string => x.toString(),
-  "number-bignumber": (x: number): BigNumber => new BigNumber(x),
+  "number-bignumber": (x: number): BigNumber => {
+    if (x < 0) {
+      throw new Error(`Input value ${x} cannot be a negative number`);
+    }
```
LayneHaber: we wouldnt want to check for this within the cast functions, those should be completely independent of validation

modules/client/src/validation/addresses.ts
```diff
@@ -15,15 +15,15 @@ function isValidAddress(value: any): boolean {
 
 export function invalidXpub(value: string): string | undefined {
```
LayneHaber: in all of these functions we should probably allow for more specific errors to be thrown, ie add a way to optionally append or prepend more contextual messages to the errors

modules/client/src/controllers/TransferController.ts
rhlsthrm: Ahh missed this one!

modules/types/src/convert.ts
rhlsthrm: I didn't like it either but I couldn't see how else to resolve https://github.com/ConnextProject/indra/issues/467 since conversion happens before validation. Should we do a pre-conversion check? Does it seem like we should at least try-catch around the bignumber cast?

modules/client/src/validation/addresses.ts
rhlsthrm: Right, that's a good idea. Separate from the scope of this PR though.

modules/types/src/convert.ts
rhlsthrm: @LayneHaber any thoughts on this?

LayneHaber: We could, but the negative number shouldn't cause an issue when creating a new `BigNumber` instance, seems like the problem comes from the `_hex` property when it's an object?

We could check if it has a `_hex` field and then use the value of that property?

```sh
Laynes-MBP:cf-core laynehaber$ node
> const eth = require("ethers");
> new eth.utils.BigNumber(-1);
BigNumber { _hex: '-0x01' }
> new eth.utils.BigNumber("-0x01")
BigNumber { _hex: '-0x01' }
> new eth.utils.BigNumber({ _hex: '-0x01' })
{ Error: invalid BigNumber value (arg="value", value={"_hex":"-0x01"}, version=4.0.39)
    at Object.throwError (/Users/laynehaber/Developer/indra/node_modules/ethers/errors.js:76:17)
    at new BigNumber (/Users/laynehaber/Developer/indra/node_modules/ethers/utils/bignumber.js:99:20)
  reason: 'invalid BigNumber value',
  code: 'INVALID_ARGUMENT',
  arg: 'value',
  value: { _hex: '-0x01' } }
```

rhlsthrm: Hmm okay well if it's not a negative number issue we should handle this differently. I don't see how this could be the _hex property though because all BigNumbers would have that. I'll look into this more.

rhlsthrm: Found the issue, there was a missing validator and the validator had the wrong logic.

PR #566, opened 2019-11-18
[daicard] Force everyone to cash out SAI 

rhlsthrm:
## The Problem
We can only support one token, so we are migrating from SAI to DAI. People that have SAI in their channels need to withdraw before using the Daicard.

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,11 +196,6 @@ daicard-proxy: $(shell find $(proxy) $(find_options))
 	docker build --file $(proxy)/daicard.io/prod.dockerfile --tag daicard_proxy:latest .
 	$(log_finish) && touch $(flags)/$@
 
-dashboard-prod: node-modules client $(shell find $(dashboard)/src $(find_options))
-	$(log_start)
-	$(docker_run) "cd modules/dashboard && npm run build"
-	$(log_finish) && touch $(flags)/$@
-
 database: node-modules $(shell find $(database) $(find_options))
 	$(log_start)
 	docker build --file $(database)/db.dockerfile --tag $(project)_database:latest $(database)
--- a/address-book.json
+++ b/address-book.json
@@ -78,7 +78,7 @@
     }
   },
   "4": {
-    "Token": {
+    "SAIToken": {
       "address": "0xFab46E002BbF0b4509813474841E0716E6730136",
       "bytecode": "0x9ee358e4863490e5417fd2a4504e5547dbaea7e77677d4c797db5be5418b8d30",
       "txHash": "0x0174127f12e41181608ba85e08ce8a1decfb4face489090c7c9aa33b1f1c5673"
@@ -151,7 +151,7 @@
     "ProxyFactory": {
       "address": "0x8eb543b35DE94B0E636402C7cA32947b22853eDF"
     },
-    "SAIToken": {
+    "Token": {
       "address": "0xaFF4481D10270F50f203E0763e2597776068CBc5"
     }
   },
--- a/address-book.json
+++ b/address-book.json
@@ -1,6 +1,9 @@
 {
   "1": {
     "Token": {
+      "address": "0x6B175474E89094C44Da98b954EedeAC495271d0F"
+    },
+    "SAIToken": {
       "address": "0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359",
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -196,11 +196,6 @@ daicard-proxy: $(shell find $(proxy) $(find_options))
 	docker build --file $(proxy)/daicard.io/prod.dockerfile --tag daicard_proxy:latest .
 	$(log_finish) && touch $(flags)/$@
 
-dashboard-prod: node-modules client $(shell find $(dashboard)/src $(find_options))
-	$(log_start)
-	$(docker_run) "cd modules/dashboard && npm run build"
-	$(log_finish) && touch $(flags)/$@
-
```
LayneHaber: was this intentional?

address-book.json
```diff
@@ -1,6 +1,9 @@
 {
   "1": {
     "Token": {
+      "address": "0x6B175474E89094C44Da98b954EedeAC495271d0F"
+    },
+    "SAIToken": {
       "address": "0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359",
       "bytecode": "0xa288eb5e119c76a1fc65ceb646ecf0a77657f66bb08038a226a37a467a43d035",
       "txHash": "0x91b318fd3498c14a1495357d6a4e895fd29c3d915ee695a6fbc8ef72908a1aa8"
```
LayneHaber: why do we have the address/bytecode/txhash for sai but we dont have it for dai (token) on mainnet?

address-book.json
```diff
@@ -147,6 +150,9 @@
     },
     "ProxyFactory": {
       "address": "0x8eb543b35DE94B0E636402C7cA32947b22853eDF"
+    },
+    "SAIToken": {
+      "address": "0xaFF4481D10270F50f203E0763e2597776068CBc5"
```
LayneHaber: why are we missing the information we have on mainnet for rinkeby?

modules/daicard/src/App.js
```diff
@@ -131,6 +134,7 @@ class App extends React.Component {
       useWalletConnext: false,
       sendScanArgs: { amount: null, recipient: null },
       redeemScanArgs: { amount: null, recipient: null },
+      saiAmount: Zero,
```
LayneHaber: we should store this as a `Currency` type

modules/daicard/src/components/settingsCard.js
```diff
@@ -42,14 +40,13 @@ const style = withStyles(theme => ({
   },
 }));
 
-export const SettingsCard = style(props => {
+export const SettingsCard = style(({ classes, setWalletConnext, getWalletConnext, store, history }) => {
```
LayneHaber: 🥇

modules/daicard/src/components/withdrawSai.js
```diff
@@ -0,0 +1,79 @@
...
+      aria-describedby="alert-dialog-description"
+    >
+      <DialogTitle id="alert-dialog-title">You have SAI in Your Channel!</DialogTitle>
+      <DialogContent dividers>
+        <Typography variant="body1">
+          We've transitioned the DaiCard from SAI (single-collateral DAI) to DAI (multi-collateral
+          DAI)! Please withdraw the SAI from your channel in order to keep using your daicard.
+        </Typography>
+        <Typography variant="h6" component="p">
+          Sai Balance: {saiBalance ? formatEther(saiBalance) : 0}
```
LayneHaber: we should display this in a way thats more consistent with how we display other balances. 

ie. if you have 2.1 SAI it should display $2.10 (maybe not the symbol, but definitely the decimal place)

address-book.json
rhlsthrm: It was already there for `Token`, we are moving `Token` (i.e. SAI) to `SAIToken` so I just changed the name of the JSON key. There's no way we can rely on the bytecode and txHash since we aren't going to be the ones that deploy these contracts.

address-book.json
rhlsthrm: I'm not inclined to think it was real information, since we aren't deploying token contracts to mainnet.

modules/daicard/src/components/withdrawSai.js
rhlsthrm: Will use currency type for this, good call.

PR #568, opened 2019-11-19
[cf-core, node, client] Unrolled deposit

### LINKED ISSUE #462: [protocol, card, client] Unrolled Deposit
Separate out the deposit commitment from the onchain deposit transaction itself.

WIP: https://github.com/counterfactual/monorepo/issues/2516

rhlsthrm:
## The Problem
Fixes #462 

## Checklist:


modules/cf-core/src/api.ts
```diff
@@ -35,11 +37,13 @@ const controllers = [
   CreateChannelController,
   DepositController,
   InstallAppInstanceController,
+  InstallBalanceRefundController,
```
snario: A better name might be "RequestDepositRightsController" or something more descriptive of what the intent is.

modules/cf-core/src/machine/types.ts
```diff
@@ -119,6 +119,7 @@ export type UninstallParams = {
   initiatorXpub: string;
   responderXpub: string;
   multisigAddress: string;
+  blockNumberToUseIfNecessary?: number;
```
snario: `ToUseIfNecessary` might be better described as `AtWhichUninstallIsCalled`

modules/cf-core/src/api.ts
rhlsthrm: Thanks, this is a great name.

modules/cf-core/src/methods/state-channel/uninstall-balance-refund/controller.ts
```diff
@@ -0,0 +1,72 @@
...
+        ...params,
+        // unused params to make types happy
+        tokenAddress: CONVENTION_FOR_ETH_TOKEN_ADDRESS,
+        amount: Zero
+      },
+      blockNumber
+    );
+
+    return {
+      multisigBalance
```
snario: Why is this the return type?

modules/cf-core/src/protocol/utils/get-outcome-increments.ts
```diff
@@ -84,14 +86,19 @@ export async function computeTokenIndexedFreeBalanceIncrements(
...
-    if (amount.gt(0)) {
+    const currentBlockNumber = await provider.getBlockNumber();
+    if (
+      amount.gt(0) ||
+      (blockNumberToUseIfNecessary &&
+        currentBlockNumber >= blockNumberToUseIfNecessary)
+    ) {
       return {
         [(appInstance.state as CoinBalanceRefundState).tokenAddress]: {
           [to]: amount
```
snario: I think this is right, but it would benefit from declaring some variables with names, or adding some comments, as it reads confusingly.

modules/cf-types/.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "printWidth": 80,
+  "trailingComma": "none"
+}
```
snario: Seems unrelated to this PR.

LayneHaber: it was to fix the linting from when we added these packages into our repo, but yes otherwise unrelated to this PR

modules/cf-core/src/methods/state-channel/uninstall-balance-refund/controller.ts
rhlsthrm: Callers might be interested in the multisig balance after the balance refund app is installed. Open to other suggestions on return values.

modules/cf-types/.prettierrc
rhlsthrm: Correct. This has been cherry-picked into other branches as well. Will leave it for simplicity.

modules/cf-core/src/methods/state-channel/uninstall-balance-refund/controller.ts
LayneHaber: Agree -- should be at least something closer to the updated free balance?

modules/client/src/connext.ts
```diff
@@ -227,6 +230,30 @@ export const connect = async (opts: ClientOptions): Promise<IConnextClient> => {
   log.debug("Registering subscriptions");
   await client.registerSubscriptions();
 
+  // TODO: should we have a flag for this?
```
LayneHaber: as in a flag as to whether this app is already installed or an event thrown or what

modules/client/src/connext.ts
```diff
@@ -235,7 +262,7 @@ export const connect = async (opts: ClientOptions): Promise<IConnextClient> => {
   // since if the hub never submits you should not continue interacting
   log.debug("Reclaiming pending async transfers");
   // no need to await this if it needs collateral
-  client.reclaimPendingAsyncTransfers();
+  await client.reclaimPendingAsyncTransfers();
```
LayneHaber: if it needs collateral why would we await these reclaims? would take forever for the client to start up

modules/client/src/connext.ts
rhlsthrm: A flag to enable async deposits to a token.

modules/client/src/connext.ts
rhlsthrm: @bohendo found out this was causing issues, but now I'm realizing maybe we shouldn't do this. Although maybe we should wait so that we don't let things get weird if you try to do lots of other things while this is happening?

PR #570, opened 2019-11-19
Dashboard updates

hthillman:
fix the broken things (token auth not working)


modules/dashboard/src/components/DebugChannel.js
```diff
@@ -85,15 +85,19 @@ const DebugChannel = props => {
   });
 
   const getChannelState = async () => {
+    console.log("TOKEN: ", token)
```
LayneHaber: maybe use `console.debug`?

LayneHaber: or remove?

modules/dashboard/src/components/DebugChannel.js
```diff
@@ -85,15 +85,19 @@ const DebugChannel = props => {
...
     });
 
-    var extractedValues = Object.values(Object.values(res)[0].response[0].value)[0];
+
+    var extractedValues = Object.values(res)[0].response;
     extractedValues.freeBalanceAppInstance.latestState.balances[0].forEach((balance)=>{
       balance.amount.readable = bigNumberify(balance.amount._hex).toString()
     })
+
+    console.log("Extracted: ", JSON.stringify(extractedValues))
```
LayneHaber: same with this log

Issue #525, opened 2019-11-05
[types] Event types


LayneHaber:
Events thrown from the client should be its own type (and easily detectable)

May involve PR to cf to make enums consts to easily extend their types

PR #573, opened 2019-11-20
[cf] 558 - Clean types

### LINKED ISSUE #558: [cf-core] Inconsistent types
Sometimes app instances/proposed app instances in the store are stored as objects, and sometimes they are stored as arrays. 

Sometimes you get `msg.data.data` sometimes you get `msg.data`

These should all be cleaned up so they are consistent.

Footnote: also investigate where it is `[null] vs []` when getting passed to fromJson to eliminate [this](https://github.com/ConnextProject/indra/blob/0dcc302f3bc3c0a1ae42312f1be43b303b8c3d3e/modules/cf-core/src/models/state-channel.ts#L538) fix

LayneHaber:
## The Problem
Lots of inconsistencies in the types and messages at the lower levels. 

## The Solution
Increase assertions in the cf tests and fix types where needed. Fixes #558 

## Checklist:


LayneHaber:
Things I would like to improve (but will do so in other PRs):
1. making all of the enums `const`s so we can easily extend events a la #525 
2. renaming `from` to `initiator` in the event emitted msgs
3. creating `*Data` types (e.g. `DepositStartedMessageData` type) so consumers of the package can use those easily

modules/cf-core/src/methods/state-channel/create/controller.ts
```diff
@@ -65,12 +65,15 @@ export default class CreateChannelController extends NodeController {
...
+    // use state channel for owners
+    const sc = await store.getStateChannel(multisigAddress);
+
     const msg: CreateChannelMessage = {
       from: publicIdentifier,
       type: NODE_EVENTS.CREATE_CHANNEL,
       data: {
         multisigAddress,
-        owners,
+        owners: sc.multisigOwners,
```
rhlsthrm: Why not use the owners that are passed in? Since this is the create channel method they should never be different and would avoid a DB read.

LayneHaber: It comes out to xpubs vs eth addresses, but i can use the utils functions for that

LayneHaber:
> Much needed clean up in here. Wasn't totally clear but will this fix the event symmetry issues?

Unfortunately, no. But I can add that in another PR. This one was already pretty hefty so I didn't want to add to it

modules/node/src/admin/admin.provider.ts
```diff
@@ -6,6 +6,7 @@ import { AdminMessagingProviderId, MessagingProviderId } from "../constants";
 import { AbstractMessagingProvider } from "../util";
 
 import { AdminService } from "./admin.service";
+import { StateChannelJSON } from "@connext/cf-core";
```
bohendo: If this is a type that's used by the world outside cf-core then we should consider moving this into cf-types or types and importing it into cf-core instead.

modules/node/src/util/cfCore.ts
```diff
@@ -7,18 +7,23 @@ export {
...
   ProposeMessage,
   RejectInstallVirtualMessage,
   RejectProposalMessage,
   UninstallMessage,
   UninstallVirtualMessage,
   UpdateStateMessage,
-  WithdrawMessage,
+  WithdrawConfirmationMessage,
+  WithdrawFailedMessage,
+  WithdrawStartedMessage,
```
bohendo: These messages should prob live in a types folder & be imported into cf-core instead of exported by it

bohendo:
Compiled my feedback + Layne's to-do-next list of further cleanup into issue #574 
What we have here looks great though so let's merge it & do further cleanup specified by #574 in another PR

modules/cf-core/src/methods/state-channel/create/controller.ts
rhlsthrm: True, forgot about that.

PR #578, opened 2019-11-25
553 add bytecode to state checkpoint

bohendo:
## The Problem

the getCreate2MultisigAddress function (+ a handful of other xkey utils) were copy/pasted from cf-core into the client & node. Messing w the multisig calculations is easier if we only need to mess w it in one place. 

## The Solution

The cf-core util copies have been exported from cf-core, removed from client/node and imported properly.

Additionally, a tricky race condition was fixed and bot tests have been beefed up to ensure that similar race conditions won't ever make it onto staging again.

A few `[]` vs `{}` issues were fixed too.

These changes are not strictly backwards incompatible but they do add new exports to cf-core so the interface has changed so we should increment the middle version number.

## Checklist:


modules/cf-core/src/machine/index.ts
```diff
@@ -15,35 +15,35 @@ import {
   UninstallParams,
   UninstallVirtualAppParams,
   UpdateParams,
-  WithdrawParams
+  WithdrawParams,
```
rhlsthrm: This should use CF's lint rules. Our lint rules are conflicting with theirs, and we want to be able to PR this stuff upstream and absorb upstream changes. I added their lint config to the modules in this commit:

https://github.com/ConnextProject/indra/pull/568/commits/b0b4ce99d1462e392792f9aff8fa8325b63071a4

It's stuck in my branch I realized. I'll cherry-pick this into staging.

modules/cf-core/src/machine/xkeys.ts
```diff
@@ -1,66 +1,43 @@
...
  * BIP-39 specifies how to convert mnemonic to/from entropy and mnemonic to seed
  * BIP-43 specifies that the first field should be purpose (i.e. "m / purpose'")
- * BIP-44 specifies that if the purpose is 44, then the format is "m / purpose' / cointype' / account' / change / index"
+ * BIP-44 specifies format of: "m/purpose'/cointype'/account'/change/index" (iff purpose is 44)
  */
 
-export function computeRandomExtendedPrvKey(): string {
-  return fromMnemonic(Wallet.createRandom().mnemonic).extendedKey;
-}
+const xkeyKthHDNode = memoize(
```
rhlsthrm: There's a different memoization package I've seen used in the repo: https://github.com/darrylhodgins/typescript-memoize

Is there a reason you didn't use this one? The decorator syntax is cleaner IMO, but personal preference I guess. If the other one is already a dependency no need to have multiple though.

modules/cf-core/src/machine/index.ts
LayneHaber: ^^ please 💯

modules/node/src/channel/channel.service.ts
```diff
@@ -264,16 +262,23 @@ export class ChannelService {
...
     state.singleAssetTwoPartyIntermediaryAgreements =
       Object.values(state.singleAssetTwoPartyIntermediaryAgreements).length === 0
         ? []
         : state.singleAssetTwoPartyIntermediaryAgreements;
 
+    state.proposedAppInstances =
+      Object.values(state.proposedAppInstances).length === 0 ? [] : state.proposedAppInstances;
+
+    state.appInstances = Object.values(state.appInstances).length === 0 ? [] : state.appInstances;
+
```
LayneHaber: Cleaner fix included in #577

modules/node/src/util/logger.ts
```diff
@@ -4,21 +4,21 @@ export class CLogger extends Logger {
   public cxt: string;
   constructor(context: string) {
     super();
-    this.cxt = context || "";
+    this.cxt = context || "?";
```
LayneHaber: why was this change made?

ops/test-bot.sh
```diff
@@ -31,72 +31,84 @@ function cleanup {
...
+  if [[ -f "$recipientStore" ]]
+  then recipientApps="`cat $recipientStore | grep channel | cut -d ":" -f3- | tr -d '\\\' | cut -c 3- | rev | cut -c 3- | rev | jq '.appInstances | length'`"
+  fi
+  echo -e "$divider";echo "Installed apps:"
+  echo "Sender: $senderApps"
+  echo "Recipient: $recipientApps"
+  if [[ "$senderApps" != "0" || "$recipientApps" != "0" ]]
+  then echo "Error: no lingering apps should be left uninstalled" && echo && exit 1
+  fi
+}
```
LayneHaber: 🥇

LayneHaber:
Also notice that we've removed [some cache-ing](https://github.com/ConnextProject/indra/blob/64fc3d02038134c99e008cb545d6c6f7d203a182/modules/cf-core/src/utils.ts#L42) from the `utils`, was that added anywhere else?

LayneHaber:
closing because of previous comment, will wait for full changes

modules/node/src/util/logger.ts
bohendo: At some point I was seeing the `can't access property cxt of undefined` error and wanted a visual indication of a missing context but I don't think this change actually helped anything..

modules/cf-core/src/machine/xkeys.ts
bohendo: I looked at it & thought it only works for class methods.. looking at it again I'm not sure if that's actually the case. If we can eliminate this new dep & just use the old one that'd def be better though

modules/node/src/util/logger.ts
rhlsthrm: I still see this sometimes too.

modules/node/src/channel/channel.service.ts
bohendo: Fix in 577 should prevent more bad values being saved to the db but this'll still damage-control the ones that are already there. Let's keep this until we can confirm the db no longer has any bad states.

modules/node/src/util/logger.ts
bohendo: Actually I just got the following logs so this `"?"` might be a clue 🔎
```
[Nest] 180   - 12/04/2019, 9:33 AM   [AuthService] Got channel 0x2D9077260075B11eA0874c629AEb3B45b2531DdA: undefined
[Nest] 180   - 12/04/2019, 9:33 AM   [AuthService] Acquiring a lock for a multisig w/out a channel: lock.acquire.0x2D9077260075B11eA0874c629AEb3B45b2531DdA +1ms
?
[Nest] 180   - 12/04/2019, 9:33 AM   [LockService] Using lock ttl of 90 seconds
```

bohendo:
Re prev message by @LayneHaber I think the tests were failing for a slightly more boring reason. Some install/uninstall tests were failing bc of multisig address mismatches and I dug **deep** looking for places where different sets of bytecode was being used and did all kinds of cleaning (primarily replaces references to `@counterfactual/*` contracts w the local ones) but none of it helped.. then I made the following change and the tests started passing
```
-    const proxyBytecode = ProxyFactory.evm.bytecode.object;
+    const proxyBytecode = Proxy.evm.bytecode.object;
```
Oops.
So I think we *could* actually merge a checkpoint branch and then branch off again to finish this feature

rhlsthrm:
Let's reopen and merge if everything is green then? @bohendo

LayneHaber:
wow..

okay, what would be the difference between a checkpoint and full featured PR at this point?

modules/node/src/channel/channel.service.ts
LayneHaber: Wait, why?

The `.toJson` operates on functions being retrieved from the db, im not sure that it would prevent anything from being saved incorrectly (except on the client side, in which case restore state should fix it)

PR #590, opened 2019-12-03
572 dashboard updates

hthillman:
## Checklist:


modules/node/src/admin/admin.provider.ts
```diff
@@ -38,10 +38,18 @@ class AdminMessaging extends AbstractMessagingProvider {
     return await this.adminService.getChannelStates(userPublicIdentifier);
```
LayneHaber: should probably rename this to `getChannelStateByUserId` or something more descriptive

modules/node/src/admin/admin.provider.ts
```diff
@@ -68,11 +76,21 @@ class AdminMessaging extends AbstractMessagingProvider {
       this.authService.useAdminToken(this.getChannelStates.bind(this)),
```
LayneHaber: also rename to indicate user id is needed

modules/node/src/admin/admin.service.ts
```diff
@@ -52,16 +54,25 @@ export class AdminService {
   /**  Get Channels by xpub */
 
   async getChannelStates(userPublicIdentifier: any): Promise<CFCoreRecord[]> {
-    // get channel by xpub
+    // get channel by xpubgetChannelStateByMultiSig
```
LayneHaber: how is this being passed in? the `userPublicIdentifier` should be typed as a string and used directly, this is a little confusing

modules/node/src/channel/channel.service.ts
```diff
@@ -279,6 +279,24 @@ export class ChannelService {
     return state;
   }
 
+  async getChannelStateByMultiSig(multisigAddress: string): Promise<any> {
+    const channel = await this.channelRepository.findByMultisigAddress(multisigAddress);
+    if (!channel) {
```
LayneHaber: why not just return null?

modules/node/src/transfer/transfer.repository.ts
```diff
@@ -32,6 +32,10 @@ export class LinkedTransferRepository extends Repository<LinkedTransfer> {
     });
   }
 
+  async findAllTransfers(available: boolean = true): Promise<LinkedTransfer[]> {
+    return await this.find({ where: { available }  });
+  }
```
LayneHaber: what is the `available` flag in the context of linked transfers?

modules/node/src/channel/channel.service.ts
```diff
@@ -279,6 +279,24 @@ export class ChannelService {
...
+    }
+    const { data: state } = await this.cfCoreService.getStateChannel(channel.multisigAddress);
+
+    // FIXME: this has to be [] not {}, we should fix this better
+    state.singleAssetTwoPartyIntermediaryAgreements =
+      Object.values(state.singleAssetTwoPartyIntermediaryAgreements).length === 0
+        ? []
+        : state.singleAssetTwoPartyIntermediaryAgreements;
+
+    return state;
```
LayneHaber: this is fixed by calling `.toJson()` on a state channel object

modules/node/src/channel/channel.service.ts
```diff
@@ -279,6 +279,24 @@ export class ChannelService {
...
+  }
+
   async getAllChannelsState(): Promise<any> {
```
LayneHaber: lets try not to add anymore returning `any` types, iirc this should be `Channel[]` type

modules/node/src/admin/admin.service.ts
```diff
@@ -52,16 +54,25 @@ export class AdminService {
...
+
   //Get all channels
   async getAllChannelsState(): Promise<CFCoreRecord[]> {
     // get channel by xpub
     return await this.channelService.getAllChannelsState();
   }
 
+  async getAllTransfers(): Promise<any[]> {
+    return await this.transferService.fetchAllTransfers();
+  }
```
LayneHaber: change typing to `Transfer[]`

modules/node/src/admin/admin.service.ts
```diff
@@ -52,16 +54,25 @@ export class AdminService {
...
-    // get channel by xpub
+    // get channel by xpubgetChannelStateByMultiSig
     return await this.channelService.getChannelState(userPublicIdentifier.id);
   }
 
+  async getChannelStateByMultiSig(multisigAddress: string): Promise<CFCoreRecord[]> {
+    // get channel by xpub
+    return await this.channelService.getChannelStateByMultiSig(multisigAddress);
+  }
+
```
LayneHaber: Tracing through the code path, this should return a `StateChannelJSON` object (once typings on the `getChannelStateByMultisig` fn in `channelService` has better typings)

modules/node/src/admin/admin.provider.ts
```diff
@@ -38,10 +38,18 @@ class AdminMessaging extends AbstractMessagingProvider {
...
+    return await this.adminService.getChannelStateByMultiSig(multisigAddress);
+  }
+
   async getAllChannelsState():  Promise<any[]> {
     return await this.adminService.getAllChannelsState();
   }
 
+  async getAllTransfers():  Promise<any[]> {
+    return await this.adminService.getAllTransfers();
+  }
```
LayneHaber: `Transfer[]`

modules/node/src/admin/admin.provider.ts
```diff
@@ -38,10 +38,18 @@ class AdminMessaging extends AbstractMessagingProvider {
     return await this.adminService.getChannelStates(userPublicIdentifier);
   }
 
+  async getChannelStateByMultiSig(multisigAddress: string):  Promise<any[]> {
+    return await this.adminService.getChannelStateByMultiSig(multisigAddress);
+  }
```
LayneHaber: `StateChannelJson` --> should also be a single object, not an array

modules/node/src/admin/admin.provider.ts
LayneHaber: not `any[]`, instead `StateChannelJson`

modules/node/src/admin/admin.service.ts
hthillman: changed to string.

what do you mean by used directly?

modules/node/src/channel/channel.service.ts
hthillman: ease of debugging, mostly. I can change this if you want but otherwise it would just fail silently and null responses would need to be handled better upstream

modules/node/src/transfer/transfer.repository.ts
hthillman: 🤷‍♂️an unnecessary condition that appears not to have broken anything. I'll confirm it works without and if so remove it

modules/node/src/channel/channel.service.ts
hthillman: This wasn't my addition...  possibly a weird merge thing?

modules/node/src/admin/admin.service.ts
hthillman: ^won't this break? if it's many transfers not one?

modules/node/src/admin/admin.provider.ts
hthillman: see above --- won't this break?

PR #593, opened 2019-12-03
[cf-core] 553 add bytecode to state

### LINKED ISSUE #553: [cf-core] Add bytecode to channel store objects
- [ ] Update NetworkContext object to accept bytecode in addition to contract addresses (alternatively, just make a separate parameter for creating CF Core object).
- [ ] Store bytecode in channel store as part of creation
- [ ] Use stored bytecode for multisig address calculations

LayneHaber:
## The Problem
Fixes #553 

May need to be broken out into smaller PRs, this is to get reviews going and keep master progress into larger PR.

## The Solution
Add byte code to the state in preparation to make changes to the `ProxyFactory` contract.

Outstanding question -- will there need to be a DB migration or tool to align existing channels with the appropriate bytecode? Could lead to some rather nasty backwards incompatibilities

## Checklist:


bohendo:
Progress Report:

Withdrawals have been **significantly** robustified, a whole category of proxy mismatch problems are no longer relevant. We used to have serious problems requiring manual deployment of the factory because the Proxy's artifact bytecode did not match the bytecode embedded into the Proxy factory. But no more! We're going `proxyFactory.proxyCreationCode()` maximalist from here on out.

The proxy artifact bytecode itself should be considered, from here on out, toxic waste. We should never import it or even look at the bytecode in those contract artifacts. The proxy bytecode should **only** ever be accessed via the proxyFactory's proxyCreationCode method because this is the only bytecode that can be used to deploy a proxy during withdrawal.

Potential problem: it's very possible that existing channels have been infected by this toxic waste aka their multisig addresses were calculated by artifact bytecode and doesn't match the embeded bytecode of any of our ProxyFactories.. Theoretically, I think this could result in irreversible loss of finds. In practice, I bet old versions of the proxyFactory will be able to deploy to all existing multisig addresses.

modules/cf-core/src/message-handling/handle-protocol-message.ts
```diff
@@ -280,15 +280,20 @@ async function getQueueNamesListByProtocolName(
...
+  } = requestHandler;
 
   async function multisigAddressFor(xpubs: string[]) {
     const allowGenerated = protocol === Protocol.Setup || protocol === Protocol.InstallVirtualApp;
     return await store.getMultisigAddressWithCounterparty(
       xpubs,
       networkContext.ProxyFactory,
       networkContext.MinimumViableMultisig,
-      allowGenerated,
+      allowGenerated ? provider : undefined
```
LayneHaber: technically, in the install virtual app protocol, it should only `allowGenerated` for the channel between the initiator and the responder, but not be allowed for the initiator and intermediary or the responder and intermediary

modules/cf-core/src/methods/state-channel/deploy-state-deposit-holder/controller.ts
```diff
@@ -21,8 +24,8 @@ import {
   NO_TRANSACTION_HASH_FOR_MULTISIG_DEPLOYMENT
 } from "../../errors";
 
-// Estimate based on:
-// https://rinkeby.etherscan.io/tx/0xaac429aac389b6fccc7702c8ad5415248a5add8e8e01a09a42c4ed9733086bec
+// Estimate based on rinkeby transaction:
+// 0xaac429aac389b6fccc7702c8ad5415248a5add8e8e01a09a42c4ed9733086bec
```
LayneHaber: why not leave in the full link?

modules/cf-core/src/methods/state-channel/deploy-state-deposit-holder/controller.ts
```diff
@@ -60,7 +63,7 @@ async function sendMultisigDeployTx(
   signer: Signer,
   stateChannel: StateChannel,
   networkContext: NetworkContext,
-  retryCount: number = 3
+  retryCount: number = 1
```
LayneHaber: we should have this parameter be configurable by the caller, or retry based on known error messages (cc @rhlsthrm)

also, we need to make sure that in the `deposit` method, if this call fails it will uninstall the `coinBalanceRefund` app within the operation. (may be out of scope of this PR)

modules/cf-core/src/models/state-channel.ts
```diff
@@ -596,6 +596,12 @@ export class StateChannel {
     };
   }
 
+  /**
+   * The state channel JSON object should *always* have an associated proxy
+   * bytecode. There is no case where a JSON version of a state channel is
+   * created that did *not* have an associated bytecode with it
+   *
+   */
```
LayneHaber: would the plan to make this backwards compatible add a check for byte code and then an insertion here? (and a corresponding assertion in the `toJson` function?

modules/cf-core/src/node.ts
```diff
@@ -43,7 +39,6 @@ export class Node {
   private readonly outgoing: EventEmitter;
 
   private readonly protocolRunner: ProtocolRunner;
-  private readonly networkContext: NetworkContext;
```
LayneHaber: is there a reason not to save the network context to the node?

LayneHaber: thinking about if there is a case where a default counterfactual context is used when we try to PR this upstream

modules/cf-core/src/node.ts
```diff
@@ -64,7 +59,7 @@ export class Node {
   static async create(
     messagingService: NodeTypes.IMessagingService,
     storeService: NodeTypes.IStoreService,
-    networkOrNetworkContext: EthereumNetworkName | NetworkContext,
+    networkContext: NetworkContext,
```
LayneHaber: again, wondering about the upstream PR default context via network name case

modules/cf-core/src/protocol/install-virtual-app.ts
```diff
@@ -1108,11 +1105,11 @@ async function getUpdatedStateChannelAndVirtualAppObjectsForInitiating(
   const initiatorAddress = xkeyKthAddress(initiatorXpub, 0);
   const intermediaryAddress = xkeyKthAddress(intermediaryXpub, 0);
 
-  const multisigAddressWithIntermediary = Store.getMultisigAddressWithCounterpartyFromMap(
+  const multisigAddressWithIntermediary = await Store.getMultisigAddressWithCounterpartyFromMap(
```
LayneHaber: we should replace the static function of the store class with a utils like one

modules/cf-core/src/protocol/install-virtual-app.ts
```diff
@@ -1171,11 +1168,10 @@ async function getUpdatedStateChannelAndVirtualAppObjectsForIntermediary(
     tokenAddress
   } = params as InstallVirtualAppProtocolParams;
 
-  const stateChannelWithAllThreeParties = getOrCreateStateChannelWithUsers(
+  const stateChannelWithAllThreeParties = await getOrCreateStateChannelWithUsers(
```
LayneHaber: note that if this function is called and the network has a provider, it may not enforce the multisig is pulled from the store

LayneHaber: it *looks* like that has been accounted for by use of the static function though, noice

modules/cf-core/src/protocol/uninstall-virtual-app.ts
```diff
@@ -549,24 +548,22 @@ async function getUpdatedStateChannelAndAppInstanceObjectsForInitiating(
     stateChannelWithIntermediary,
     stateChannelWithResponding
   ] = [
-    getStateChannelFromMapWithOwners(
+    await getStateChannelFromMapWithOwners(
```
LayneHaber: same case as install virtual here, because the `getStateChannelFromMapWithOwners` function provides the network as a param, if the `ProxyFactory` bytecode changes in the supplied network context the state channel it pulls from the map could be `undefined` (when we very much expect it to be defined).

the static function should be used in cases where the channel is expected to be properly stored (ie any channel with the intermediary)

modules/cf-core/src/request-handler.ts
```diff
@@ -133,16 +133,7 @@ export class RequestHandler {
...
-      const signer = await (this.provider as JsonRpcProvider).getSigner();
-      await signer.getAddress();
-      return signer;
-    } catch (e) {
-      if (e.code === "UNSUPPORTED_OPERATION") {
-        return this.wallet;
-      }
-      throw Error(prettyPrintObject(e));
-    }
+    return this.wallet;
```
LayneHaber: this shouldn't be async anymore

modules/cf-core/src/store.ts
```diff
@@ -30,6 +27,49 @@ export class Store {
     private readonly storeKeyPrefix: string
   ) {}
 
+  // TODO: remove if store is added to Context type
+  public static async getMultisigAddressWithCounterpartyFromMap(
```
LayneHaber: make this a non-static util function

also, what would be the motivation for adding the store to the `Context` type?

modules/cf-core/test/contracts.ts
```diff
@@ -0,0 +1,163 @@
...
+    wallet,
+  ).deploy();
+
+  const singleAssetTwoPartyCoinTransferInterpreter = await new ContractFactory(
+    SingleAssetTwoPartyCoinTransferInterpreter.abi,
+    SingleAssetTwoPartyCoinTransferInterpreter.evm.bytecode,
+    wallet,
+  ).deploy();
+
+  return {
```
LayneHaber: where does the provider get added in the tests now?

modules/cf-core/test/global-setup.jest.ts
```diff
@@ -1,20 +1,44 @@
...
-  C_EXTENDED_PRIVATE_KEY
+  C_EXTENDED_PRIVATE_KEY,
 } from "./test-constants.jest";
 
 dotenvExtended.load();
+const env = {
+  ETHPROVIDER_URL: process.env.ETHPROVIDER_URL || "http://localhost:8545",
+  SUGAR_DADDY:
+    process.env.SUGAR_DADDY ||
+    "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat",
```
LayneHaber: its an env var now 🤣

modules/cf-core/test/global-teardown.jest.ts
```diff
@@ -1,3 +1 @@
-export default async function globalTeardown() {
-  global["chain"].server.close();
-}
+export default async function globalTeardown(): Promise<void> {}
```
LayneHaber: will these testing infra changes cause problems in upstream CF PRs?

modules/cf-core/src/methods/state-channel/deploy-state-deposit-holder/controller.ts
bohendo: linter was complaining about line being too long
I'm a slave to the linter

modules/cf-core/src/machine/protocol-runner.ts
```diff
@@ -79,6 +80,7 @@ export class ProtocolRunner {
     public readonly network: NetworkContext,
     public readonly provider: BaseProvider
   ) {
+    this.network.provider = network.provider || provider;
```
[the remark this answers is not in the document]
bohendo: bc provider is needed in getCreate2Address (to fetch the embedded bytecode) & the network context is already injected everywhere and I got sick of injecting the provider everywhere too.. also the provider kinda feels like part of the network context to me

modules/cf-core/src/methods/state-channel/deploy-state-deposit-holder/controller.ts
bohendo: 100% down to retry dependending on error messages.
That said, having the default be anything other than 1 is, in my humble opinion, complete madness.

modules/cf-core/src/protocol/install-virtual-app.ts
bohendo: I'm onboard w this

modules/cf-core/test/global-teardown.jest.ts
bohendo: They're small changes to just a couple files so merging these changes should be painless.

modules/cf-core/src/methods/state-channel/deploy-state-deposit-holder/controller.ts
LayneHaber: made it an issue -- #599

modules/cf-core/src/models/state-channel.ts
bohendo: No assertion.
Also this comment is out of date, we don't directly care about the proxy bytecode anymore, we care about the factory address (from which we can retrieve the *real* bytecode).
But if no address exists idk, prob try to use the current one from NetworkContext and throw if that one doesn't match

modules/cf-core/src/node.ts
bohendo: Default context was kinda useless imo (hence why I killed it).
It wasn't being used anywhere and contract addresses are too important to just quietly fall back on defaults that are probably out of date. I'd rather be clear: either the layer above gives us addresses that they want to use, or this layer should freak out.

modules/cf-core/src/store.ts
bohendo: Idk the motivation, I think cf did that in the store refactor that we avoided pulling in (I just moved this fn to the top so it looks like I wrote this)

modules/cf-core/test/contracts.ts
bohendo: Tests get an `ETHPROVIDER_URL` env var now
But tests were already using: `global["ganacheUrl"] = env.ETHPROVIDER_URL;` that earlier pointed to the local ganache so not much changed outside of the test setup.

bohendo: Tests already used `global["ganacheUrl"]` to setup the provider. In global-setup we point to an external ganache instead of the local one they used to use but besides that nothing else in tests really needed to change.

PR #598, opened 2019-12-06
[cf-core, client, node, types] 553 next steps

### LINKED ISSUE #569: Upgrade CI
2 things stand out as slowing down the CI process:

#### Soljs

We compile solidity from 4 modules (`contracts`, 'cf-apps', and funding + adjudicator contracts) using soljs which is way slower than solc compiled from c. We upgrade builder to include a solc binary & configure our solidity modules waffle.json files to use that instead of soljs.

#### We `make prod` too much

We `make prod` before each of our 4 set of tests and then again before `make push` in staging/prod CI. The full build before each set of tests isn't actually a huge deal bc those can run in parallel but that still clogs up our CI. Then we serially `make prod` again before pushing though, that's no bueno. Find a better way to `make prod` once and then not again during same CI run.

Saving testing data like docker images & build artifacts between jobs is complicated & vendor-specific in CircleCI. That kind of cross-job data storage might be inherently vendor-specific though. If we gotta do something vendor-specific, might as well use GitHub's CI since that'll keep everything in one place.

LayneHaber:
## The Problem
Just because the battle has been won against #553 with #593 does not mean the war is over.

## The Solution
Finishing off the issues by:
- [x] Adding `ProxyFactory` ~bytecode~ address to each newly created state channel.
- [x] Proactively ensuring the `multisigAddress` is correctly generated by the `ProxyFactory` address in the state before deploying the state deposit holder address and before any funding protocols are completed (`deposit`, `requestDepositRights`).
- [x] Adding an admin endpoint to retroactively update all channels with correct `ProxyFactory` addresses in their store

Additional context on "Adding an admin endpoint to retroactively update all channels with correct `ProxyFactory` addresses in their store"
> admin endpoint will try to verify (state): boolean ie true if the state's proxyFactoryAddress (or fallback on the current one in networkContext if does not exist yet) has a proxyCreateCode() that calculates to the expected multisig address
> If not, we need to (manually?) check the historical proxyFactoryAddresses to see if any have embedded createCode to yield the expected multisig
> else: oh no :fire: :dollar: :fire:

Edit 12/9 EOD:

closes #569 

Edit 12/9:
Adding historical proxy factory address that we have seen:

### Rinkeby
- `0x49f2eCa045B2372C334B6CcBB9232C48f9acA097`
- `0x8A49B435cc3D2176B67e0D26170387EeDf135669`
- `0xD891F41c4ba30b1FF4f604e30F64ae387DD85b4F`
- `0x6CF0c4Ab3F1e66913c0983DC0bb1202d958ABb8f`
- `0xc8d7Cf5638dfa5b79A070c5aC983716575bEF4B0`
- `0xc40E9B210363163a143f454fa505ACeAA28Cd475`
- `0x8eb543b35DE94B0E636402C7cA32947b22853eDF`
- `0xc8d7Cf5638dfa5b79A070c5aC983716575bEF4B0`
- `0x8eb543b35DE94B0E636402C7cA32947b22853eDF`

### Mainnet
- [`0x90Bf287B6870A99E32130CED0Da8b02302a8a4dE`](https://github.com/ConnextProject/indra/commit/097802b62863a88f2c4e9c8c639634e93dcfd364#diff-1d94c5cc1f532e49a22ee6fd8cae03c3)
- [`0xA16d9511C743d6D6177A65892DC2Eafd417BfD7A`](https://github.com/ConnextProject/indra/commit/d08b0aa9bbbd464b157bf6f3e332a9b33a9b0d6b#diff-1d94c5cc1f532e49a22ee6fd8cae03c3)
- [`0xc756Bf6A685573C6879D4363401940f02B4E27a1 `](https://github.com/ConnextProject/indra/commit/a78bd617ee806e419e7b108cf539df276c82f6a5#diff-1d94c5cc1f532e49a22ee6fd8cae03c3)
- [`0x6CF0c4Ab3F1e66913c0983DC0bb1202d958ABb8f `](https://github.com/ConnextProject/indra/commit/475d6dca9c6daf2b1d504b59dd0e0c12d9922a8c#diff-1d94c5cc1f532e49a22ee6fd8cae03c3)
- [`0x711C655e08aaA9081e0BDc431920507CCD96b7a0`](https://github.com/ConnextProject/indra/commit/7721cdeb867d9e2a828f39d1ae4eb1c932d2c3b7#diff-1d94c5cc1f532e49a22ee6fd8cae03c3)
- [`0xF9015aA98BeBaE3e43945c48dc3fB6c0a5281986`](https://github.com/ConnextProject/indra/commit/c0e0de60abebdfc7555b5dba4111be21c7ec670f#diff-1d94c5cc1f532e49a22ee6fd8cae03c3)

## Checklist:


rhlsthrm:
@LayneHaber updated the description to make progress/status a little more accurate.

rhlsthrm:
> Proactively ensuring the multisigAddress is correctly generated by the ProxyFactory address in the state before deploying the state deposit holder address and before any funding protocols are completed (deposit, requestDepositRights).

Completed this task with the last few commits. Would like @bohendo and @LayneHaber to review.

modules/cf-core/src/methods/state-channel/deposit/controller.ts
```diff
@@ -47,6 +49,17 @@ export default class DepositController extends NodeController {
 
     const channel = await store.getStateChannel(multisigAddress);
 
+    const expectedMultisigAddress = await getCreate2MultisigAddress(
+      channel.userNeuteredExtendedKeys,
+      networkContext.ProxyFactory,
+      networkContext.MinimumViableMultisig,
+      provider
+    );
```
LayneHaber: you need to use the channels proxy factory address here, not the one from the `networkContext`

`networkContext` could change during the lifetime of the channel, but using the `proxyFactoryAddress` that's stored in the channel should always work.

modules/cf-core/src/methods/state-channel/request-deposit-rights/controller.ts
```diff
@@ -29,7 +31,26 @@ export default class RequestDepositRightsController extends NodeController {
...
+    const { multisigAddress } = params;
+
+    const channel = await store.getStateChannel(multisigAddress);
+
+    const expectedMultisigAddress = await getCreate2MultisigAddress(
+      channel.userNeuteredExtendedKeys,
+      networkContext.ProxyFactory,
+      networkContext.MinimumViableMultisig,
+      provider
+    );
```
LayneHaber: same as above, must be channel's `proxyFactoryAddress`

modules/cf-core/src/methods/state-channel/withdraw-commitment/controller.ts
```diff
@@ -20,6 +22,27 @@ export default class WithdrawCommitmentController extends NodeController {
...
+    const { multisigAddress } = params;
+
+    const channel = await store.getStateChannel(multisigAddress);
+
+    const expectedMultisigAddress = await getCreate2MultisigAddress(
+      channel.userNeuteredExtendedKeys,
+      networkContext.ProxyFactory,
+      networkContext.MinimumViableMultisig,
+      provider
+    );
```
LayneHaber: ditto

modules/cf-core/src/protocol/propose.ts
```diff
@@ -44,7 +44,7 @@ export const PROPOSE_PROTOCOL: ProtocolExecutionFlow = {
 
     const preProtocolStateChannel = stateChannelsMap.get(multisigAddress)
       ? stateChannelsMap.get(multisigAddress)!
-      : StateChannel.createEmptyChannel(multisigAddress, [
+      : StateChannel.createEmptyChannel(multisigAddress, network.ProxyFactory, [
```
LayneHaber: note: this is done to allow for virtual proposals, but it still makes me a *bit* uneasy..

modules/node/src/admin/admin.provider.ts
```diff
@@ -39,7 +39,7 @@ class AdminMessaging extends AbstractMessagingProvider {
     return await this.adminService.getChannelState(userPublicIdentifier);
   }
 
-  async getAllChannelsState():  Promise<any[]> {
+  async getAllChannelsState(): Promise<any[]> {
```
LayneHaber: boo `any[]` think i fixed this in #570

modules/node/src/admin/admin.service.ts
```diff
@@ -177,4 +179,60 @@ export class AdminService {
...
+  async getIncorrectProxyFactoryAddresses(): Promise<string[]> {
+    const channels = await this.channelService.findAll();
+
+    const incorrectChannels = [];
+    for (const channel of channels) {
+      const { data: stateChannel } = await this.cfCoreService.getStateChannel(
+        channel.multisigAddress,
+      );
+      console.log("stateChannel: ", stateChannel);
+      const contractAddresses = await this.configService.getContractAddresses();
```
LayneHaber: there have been a couple proxy factories so far, we should be fetching all of them

even worse, we were using the `bytecode` in the deps over the contract method iirc, so we will have to make sure the all the used `bytecode` lines up with what's onchain

modules/node/src/admin/admin.service.ts
```diff
@@ -177,4 +179,60 @@ export class AdminService {
...
+        stateChannel.proxyFactoryAddress || contractAddresses.ProxyFactory,
+        contractAddresses.MinimumViableMultisig,
+        this.configService.getEthProvider(),
+      );
+
+      if (expectedMultisigAddress !== stateChannel.multisigAddress) {
+        logger.error(
+          `expectedMultisigAddress ${expectedMultisigAddress} not equal to stateChannel.multisigAddress ${stateChannel.multisigAddress}`,
+        );
+        incorrectChannels.push(stateChannel.multisigAddress);
```
LayneHaber: we should include the `bytecode` or factory address we found that *correctly* generates the multisig

modules/node/test-nats.ts
```diff
@@ -9,7 +9,7 @@ async function start(): Promise<void> {
   // );
   // console.log("helloMsg: ", helloMsg);
   const history = await nc.request(
-    "transfer.get-history.xpub6DXwZMmWUq4bRZ3LtaBYwu47XV4Td19pnngok2Y7DnRzcCJSKCmD1AcLJDbZZf5dzZpvHqYzmRaKf7Gd2MV9qDvWwwN7VpBPNXQCZCbfyoK",
+    "admin.get-incorrect-proxy-address",
```
LayneHaber: i completely forgot about this file, this is so much better than what i was doing 🤦‍♀

modules/node/src/admin/admin.service.ts
LayneHaber: wrong place for this now that im looking at the whole file, my b

modules/cf-core/src/methods/state-channel/deposit/controller.ts
rhlsthrm: Ahh right that makes sense!

modules/node/src/admin/admin.service.ts
rhlsthrm: I thought we changed this to be using the contract method. `bytecode` is toxic waste. Where are we storing all the historical proxy factory addresses?

modules/node/src/admin/admin.provider.ts
```diff
@@ -129,6 +133,11 @@ class AdminMessaging extends AbstractMessagingProvider {
       "admin.get-channels-for-merging",
       this.authService.useAdminToken(this.getChannelsForMerging.bind(this)),
     );
+
+    await super.connectRequestReponse(
+      "admin.fix-proxy-addresses",
+      this.authService.useAdminToken(this.fixProxyFactoryAddresses.bind(this)),
+    );
```
LayneHaber: my guess is well want to do a dry run here? or at least also have access to the `get-bad-proxy` addresses functionality

modules/node/src/admin/admin.service.ts
```diff
@@ -1,21 +1,52 @@
...
+    "0x8A49B435cc3D2176B67e0D26170387EeDf135669",
+    "0xD891F41c4ba30b1FF4f604e30F64ae387DD85b4F",
+    "0x6CF0c4Ab3F1e66913c0983DC0bb1202d958ABb8f",
+    "0xc8d7Cf5638dfa5b79A070c5aC983716575bEF4B0",
+    "0xc40E9B210363163a143f454fa505ACeAA28Cd475",
+    "0x8eb543b35DE94B0E636402C7cA32947b22853eDF",
+    "0xc8d7Cf5638dfa5b79A070c5aC983716575bEF4B0",
+    "0x8eb543b35DE94B0E636402C7cA32947b22853eDF",
+  ],
+};
```
LayneHaber: noice, wondering if there's a good way to keep this up to date in an automated way?

modules/node/src/admin/admin.service.ts
```diff
@@ -202,4 +233,106 @@ export class AdminService {
...
+      );
+      const contractAddresses = await this.configService.getContractAddresses();
+
+      if (stateChannel.multisigAddress !== channel.multisigAddress) {
+        // things will be very weird if this happens, so just dont allow this to happen
+        logger.error(
+          `This should never happen! stateChannel.multisigAddress ${stateChannel.multisigAddress} !== channel.multisigAddress ${channel.multisigAddress}`,
+        );
+        continue;
+      }
```
LayneHaber: wait -- this does happen in some cases where the multisig address changed and nothing stopped it from getting written to the store.

this is actually the exact check i used manually to verify *if* the multisig address changed over the lifetime of the channel

rhlsthrm: Well we should have fixed all of them by now, so if we didn't let's blow up and fix it first!

Issue #583, opened 2019-11-26
[refactor] Accept only ChannelProvider objects


pedrouid:
## Introduction
Right now, the client accepts three ways to be instantiated. It could either be through a `mnemonic`, `xpub + keygen` or `ChannelProvider`. The first two are essentially the same as on connect the `mnemonic` is used to generate an `xpub + keygen` to be passed to the CFCore module. 

In the current codebase a lot of the methods include switch statements to handle RPC methods to be called using either the ChannelProvider or CounterfactualNode. Hence it would be a better practice to abstract the CounterfactualNode as a ChannelProvider that accesses CFCore module. This why we could remove all switch statements and remove the RpcType from the client. 

Additionally, the Connext client connect method always calls the ChannelProvider enable method and it would be better practice to do this externally.

## ClientOptions

Using the WalletConnect ChannelProvider the client options would be as follows
```typescript
const channelProvider = new WalletConnectChannelProvider({ ...walletConnectConfig });

const options: ClientOptions = {
  ethProviderUrl: `https://daicard.io/api/ethprovider`,
  channelProvider,
}

const channel: Channel = await connext.connect(options)
```

Therefore the refactor would move the CounterfactualNode logic from the client's channelRouter to a channelProvider as follows

**Before**
```typescript
const options: ClientOptions = {
  mnemonic: 'Apple Banana ...',
  ethProviderUrl: `https://daicard.io/api/ethprovider`,
  nodeUrl: `wss://daicard.io/api/messaging`,
  store
}
```

**After**
```typescript
const channelProvider = new CounterfactualNodeChannelProvider({
  mnemonic: 'Apple Banana ...',
  store
});

const options: ClientOptions = {
  ethProviderUrl: `https://daicard.io/api/ethprovider`,
  nodeUrl: `wss://daicard.io/api/messaging`,
}

const channel: Channel = await connext.connect(options)
```

This not only simplify the connext client logic but have better seperation of concerns focusing on only connext exclusive logic inside the client without requiring unnecessary configuration like menmonic and store which are not used by other channel providers like WalletConnect for example.

Thus the ClientOptions will only require connext specific parameters like `ethProviderUrl` and `nodeUrl`.

## CounterfactualNode ChannelProvider
As also described in the introduction the two possible configuration options for CounterfactualNode ChannelProvider would be `mnemonic` and `xpub + keygen`. 

```typescript
// using mnemonic
const channelProvider = new CounterfactualNodeChannelProvider({
[...]

PR #608, opened 2019-12-10
Channel Provider refactor

pedrouid:
## Changelog
* Refactor ChannelRouter to only accept JSON-RPC
* Clean up WalletConnect client mapping
* Add new JSON-RPC methods
* Allow dumb clients to make store requests to smart clients

## Context
The main motivation for this PR was to solve issues that I found on both Torus and WalletConnect channelProviders as they were both having issues with the client mapping. The way it worked before basically duplicated work between dumb and smart clients yet it was simply necessary to directly hit the channelRouter. 

So all incoming JSON-RPC requests from transport don't need to be mapped individually and can be passed directly as is. This will simplify channelProviders in the future and it also basically converts the current channelRouter as a Counterfactual ChannelProvider.

The main difference this PR introduces is that channelRouter methods are now accepted as JSON-RPC methods only


pedrouid:
Just renamed all channelRouter to channelProvider

pedrouid:
Fix #583

pedrouid:
Replaced by #612

PR #612, opened 2019-12-10
Channel Provider Refactor

### LINKED ISSUE #583: [refactor] Accept only ChannelProvider objects

pedrouid:
## Changelog
* Refactor ChannelRouter to only accept JSON-RPC
* Clean up WalletConnect client mapping
* Add new JSON-RPC methods
* Allow dumb clients to make store requests to smart clients
* Rename ChannelRouter to ChannelProvider

## Context
The main motivation for this PR was to solve issues that I found on both Torus and WalletConnect channelProviders as they were both having issues with the client mapping. The way it worked before basically duplicated work between dumb and smart clients yet it was simply necessary to directly hit the channelRouter. 

So all incoming JSON-RPC requests from transport don't need to be mapped individually and can be passed directly as is. This will simplify channelProviders in the future and it also basically converts the current channelRouter as a Counterfactual ChannelProvider.

The main difference this PR introduces is that channelRouter methods are now accepted as JSON-RPC methods only

## Related Issues

Fix #583


bohendo:
Closed in favor of PR #638

Issue #574, opened 2019-11-23
Polish cf-core types


bohendo:
Various cleanup chores surrounding types. If any of these are big tasks on their own then they should be broken out into a separate issue but for now we'll group them all here.

 - [x] The StateChannelJSON type should live in a types module (types or cf-types) and be imported by cf-core instead of exported by it.
 - [ ] Above but for all those message types eg `WithdrawFailedMessage`.
 - [ ] making all of the enums consts so we can easily extend events a la #525
 - [ ] renaming from to initiator in the event emitted msgs
 - [ ] creating *Data types (e.g. DepositStartedMessageData type) so consumers of the package can use those easily
 - [ ] symmetric event data emission
 - [ ] verify `result` returned from the rpc methods 

See comments on PR #573 for more info

LayneHaber:
We should probably break these out into separate issues, but keep this one as a reference check list

LayneHaber:
Fixed event enums in #613 PR, still need to fix other miscellaneous enums

LayneHaber:
Closing, broke this out into different issues:

- #623 making all of the `enum`s `const`s
- #624 event messages in types modules
- #625 rename `from` to `initiator`
- #626 creating `*Data` types
- #627 symmetric event emissions
- #628 rpc result assertion

PR #613, opened 2019-12-10
[cf-core, cf-types, node, client] 525 event types

### LINKED ISSUE #525: [types] Event types

LayneHaber:
## The Problem
Typescript `enum`s are not easily extensible, which is annoying.

Fixes #525 and partially addresses #574 

## The Solution
Convert the `enum EventName` to a `const`.

Most of the file changes are to fix the build, the important bits are in the cf-types package [here](https://github.com/ConnextProject/indra/blob/3cff8138b008286f0f187f87db651c916dada9f8/modules/cf-types/src/node.ts#L120) and the types package [here](https://github.com/ConnextProject/indra/blob/3cff8138b008286f0f187f87db651c916dada9f8/modules/types/src/basic.ts#L9).

## Checklist:


modules/cf-core/src/message-handling/handle-protocol-message.ts
```diff
@@ -177,14 +178,14 @@ async function getOutgoingEventDataFromProtocol(
       // determine if the withdraw is finishing or if it is starting
       return {
         ...baseEvent,
-        type: NODE_EVENTS.WITHDRAWAL_STARTED,
+        type: "WITHDRAWAL_STARTED_EVENT",
```
rhlsthrm: How come this one didn't need to be downcasted?

modules/cf-core/src/node.ts
```diff
@@ -265,7 +265,10 @@ export class Node {
    * @param event
    * @param callback
    */
-  on(event: string, callback: (res: any) => void) {
+  on(
+    event: NodeTypes.EventName | NodeTypes.RpcMethodName,
```
rhlsthrm: 👍

rhlsthrm: Also, it's not really making sense that you can subscribe to both `NodeTypes.EventName` and `NodeTypes.RpcMethodName`. Should we consolidate all events you can subscribe to into `NodeTypes.EventName`?

PR #618, opened 2019-12-11
Connect Refactor

### LINKED ISSUE #583: [refactor] Accept only ChannelProvider objects

### LINKED ISSUE #637: [client] Connect refactor
A separate refactor that inherits from ChannelProvider refactor described in #583.

The motivation behind this one is to separate concerns better in the connect method to solve the multisigAddress assignment eventually.

Also it allows all ClientOptions to be optional except for ethProviderUrl. A connext client instantiated with channelProvider should only need { ethProviderUrl, channelProvider }. The interface should be improved to account for the possible combinations:

```typescript
// dumb client w/ channelProvider
clientOptions = {
  ethProviderUrl,
  channelProvider
}

// smart client w/ mnemonic
clientOptions = {
  ethProviderUrl,
  store,
  nodeUrl,
  mnemonic
}

// smart client w/ xpub+keyGen
clientOptions = {
  ethProviderUrl,
  store,
  nodeUrl,
  xpub,
  keyGen
}
```
This refactor is meant to be backwards-compatible!

pedrouid:
closes #637 (<- PR description moved here)
closes #583

### DIFF
--- a/modules/client/src/channelProvider.ts
+++ b/modules/client/src/channelProvider.ts
@@ -2,14 +2,51 @@ import { Wallet } from "ethers";
 import { arrayify } from "ethers/utils";
 import { RpcParameters } from "rpc-server";
 
-import { withdrawalKey } from "./lib";
+import { CFCore, withdrawalKey, xpubToAddress } from "./lib";
 import {
+  CFChannelProviderOptions,
   CFCoreTypes,
   ChannelProviderConfig,
   NewRpcMethodName,
   RpcConnection,
   Store,
 } from "./types";
+
+export const createCFChannelProvider = async ({
+  ethProvider,
+  keyGen,
+  lockService,
+  messaging,
+  networkContext,
+  nodeConfig,
+  nodeUrl,
+  store,
+  xpub,
+}: CFChannelProviderOptions): Promise<ChannelProvider> => {
+  const cfCore = await CFCore.create(
+    messaging as any,
+    store,
+    networkContext,
+    nodeConfig,
+    ethProvider,
+    lockService,
+    xpub,
+    keyGen,
+  );
+  const channelProviderConfig: ChannelProviderConfig = {
+    freeBalanceAddress: xpubToAddress(xpub),
+    nodeUrl,
+    signerAddress: xpubToAddress(xpub),
+    userPublicIdentifier: xpub,
+  };
+  const channelProvider = new ChannelProvider(
+    cfCore,
+    channelProviderConfig,
+    store,
+    await keyGen("0"),
+  );
+  return channelProvider;
+};
 
 export class ChannelProvider {
   private connection: RpcConnection;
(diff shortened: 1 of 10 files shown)


bohendo:
Closed in favor of PR #638

bohendo:
Closed in favor of PR #638

PR #633, opened 2019-12-12
631 apps for channel

### LINKED ISSUE #631: [cf-core, client, node] Get all app instances for channel
getting all app instances returns all the app instances for all your channels.

add optional multisig address to get app instances for only that channel if the multisig address is provided.

may need similar changes for `getProposedApps`

LayneHaber:
## The Problem
Get apps by default returned all apps for all channels.

## The Solution
Fixes #631 

Now only returns apps for a specific channel when multisig address is provided to the params.

## Checklist:


modules/cf-core/src/store.ts
```diff
@@ -162,13 +162,37 @@ export class Store {
...
+      ? [await this.getStateChannel(multisigAddress)]
+      : [...(await this.getStateChannelsMap()).values()];
+    return chanArray.reduce(
       (lst, sc) => [...lst, ...sc.proposedAppInstances.values()],
       [] as AppInstanceProposal[]
     );
   }
 
+  /**
+   * Returns a list of proposed `AppInstanceJson`s.
```
rhlsthrm: Inaccurate comment.

modules/cf-core/src/store.ts
```diff
@@ -162,13 +162,37 @@ export class Store {
   /**
    * Returns a list of proposed `AppInstanceProposals`s.
    */
-  public async getProposedAppInstances(): Promise<AppInstanceProposal[]> {
-    return [...(await this.getStateChannelsMap()).values()].reduce(
+  public async getProposedAppInstances(
+    multisigAddress?: string
+  ): Promise<AppInstanceProposal[]> {
+    const chanArray = multisigAddress
```
rhlsthrm: When would this call not have a multisig address?

modules/node/src/cfCore/cfCore.service.ts
```diff
@@ -357,9 +359,10 @@ export class CFCoreService {
   }
 
   async getCoinBalanceRefundApp(
+    multisigAddress: string,
     tokenAddress: string = AddressZero,
   ): Promise<AppInstanceJson | undefined> {
-    const appInstances = await this.getAppInstances();
+    const appInstances = await this.getAppInstances(multisigAddress);
```
rhlsthrm: Wow, good stuff. This would have been really bad.

PR #638, opened 2019-12-12
ChannelProvider + connect refactor

### LINKED ISSUE #637: [client] Connect refactor

bohendo:
closes #612 
closes #637  

## The Problem

Messy

## The Solution

Clean it up, see issues for details

## Backwards-incompatibility

None

## Special Attention

* connect method moved to `connect.ts`
* `channelRouter.ts` replaced by `channelProvider.ts`
* channelProvider only accepts JSON-RPC requests through `send` method
* NodeAPIClient refactor how `getAuthToken` works

## Checklist:


modules/client/src/connect.ts
```diff
@@ -0,0 +1,294 @@
...
+  return { messaging, node, config };
+};
+
+const setupChannelProvider = async (
+  ethProvider: providers.JsonRpcProvider,
+  log: Logger,
+  logLevel: number,
+  providedChannelProvider?: ChannelProvider,
+  channelProviderOptions?: ChannelProviderOptions,
+): Promise<ChannelProviderSetup> => {
```
bohendo: My rule of thumb: don't save anything to a variable that we only use once; just use it in-place.

This connect function was kind of a monolith so there's a valid argument that breaking it out into bite-size pieces increases readability.

But counterargument would be that I can't read the connect function straight-through anymore. To understand what it's doing, I have to constantly track down the extracted functions (+ the extra types that were added) which is harder for me to read than one big function where everything I need is in once place.

(Obviously, if there's logic we need to re-use, then not repeating ourselves trumps any readability concerns)

I don't have super strong feelings here though, feel free to leave this as-is if you like this style better.

pedrouid: This will most likely to change and it was also mostly driven to control the scope of some variables which were shared across the whole connect method. Also the idea is to port some of this logic into the channelProvider inside the enable method.

pedrouid:
Good to merge cc @LayneHaber @ArjunBhuptani @bohendo @rhlsthrm

modules/client/src/connect.ts
```diff
@@ -0,0 +1,243 @@
...
+  let node: NodeApiClient;
+  let config: GetConfigResponse;
+
+  // setup channelProvider
+  let channelProvider: ChannelProvider;
+
+  if (providedChannelProvider) {
+    channelProvider = providedChannelProvider;
+    if (!exists(channelProvider.config)) {
+      await channelProvider.enable();
```
pedrouid: This could be replace by throwing an error as well but it wouldn't be backwards-compatible however channelProvider has only been implemented in the daicard which this change is included in the PR (https://github.com/ConnextProject/indra/pull/638/files#diff-e5c0bb91d250a5754dc0e76b4c9fa552R279)

PR #645, opened 2019-12-16
623 rpc enums to const

### LINKED ISSUE #623: [types, cf-core] Make RPC Methods enums consts
**Describe the bug**
`enum` types are hard to extend, much better to use `const` types.

Ideally would replace all `enum` types in the cf code to use `const`s, but at the least the `RPC Methods` should use `const`s.

ArjunBhuptani:
Built, but haven't run tests yet

closes #623


LayneHaber:
Also completely breaks wallet connect (which used this enum), so will need to fix that before merging

LayneHaber:
should have addressed changes, tested locally with daicard UI and the wallet connect. seems to be broken, but likely not due to changes in this PR (not sure if it was ever working?).

imo we should merge this, then address the resulting changes in #638 and make sure we fix all the flows in the provider there

PR #647, opened 2019-12-16
609 dashboard updates

hthillman:
## Checklist:


modules/dashboard/src/components/Admin/Admin.js
```diff
@@ -1,14 +1,122 @@
...
 
 const RootGrid = styled(Grid)({
   flexGrow: 1,
   padding: "5%",
 });
 
+/*
+ * Notes re ProxyBytecode
+ * the first bytecode is "", this is important bc this signals that we should use the one from the factory
+ */
```
LayneHaber: can you elaborate? all of the proxy byte code on the hub side should be fixed except for these channels:

```0x1508eCF431F5DeF63d708fa657a9c8dB5d153a78
0xf6393A66a5A667D729b2b287117055466EDdDfBb
0x99A4c63cFf0Dcd50eeCF7B867C8a7aC89696A997
0xB559B1cad8eE3ea676fFC2eBA0a115D3BACD2b67
0x60EBf23291F0a0Ad51b01f997E37441217D24d7D```

File: modules/dashboard/src/components/Admin/Admin.js [hunk 2]
```diff
@@ -1,14 +1,122 @@
...
+            ],
+          ),
+          solidityKeccak256(
+            ["bytes", "uint256"],
+            [`0x${proxyBytecode.replace("0x", "")}`, minimumViableMultisigAddress],
+          ),
+        ],
+      ).slice(-40),
+    );
+  }
```
LayneHaber: there should be utility functions for this in the cf modules that you can/should use

modules/dashboard/src/components/Admin/Admin.js
```diff
@@ -1,14 +1,122 @@
...
+              console.log(`factory:    ${factory}`);
+              console.log(`bytecode:   ${bytecode}`);
+              return [legacy, multisig, factory, bytecode];
+            }
+          }
+        }
+      }
+    }
+    return "oh no none match :("
+  }
```
LayneHaber: why not use the same functions/logic as is in the hub? those can easily be converted into admin endpoints without having to recreate all the logic here. Check out the last couple of purpose built functions [here](https://github.com/ConnextProject/indra/blob/staging/modules/node/src/admin/admin.service.ts#L267)

LayneHaber:
It could be that im not fully understanding what some of these changes are doing for the dashboard, but I don't think that logic like this should live on the frontend (and also it seems like there is duplicate logic between the dashboard and the admin endpoints)

hthillman:
That was bo and/or rahul, I’m not sure what those changes are doing

LayneHaber:
> That was bo and/or rahul, I’m not sure what those changes are doing

@bohendo @rhlsthrm any context im missing?

bohendo:
I added all that multisig scanner stuff. It was designed to be an in-depth scan for one multisig at a time so I didn't need the node's db. The tweak check way *way* faster client side so that's what I went with.

This copy of getCreate2Address is way more powerful than what's in cf-core utils, we might want to consider replacing that one with this one.

This copy of the multisig scanner also scans way more combinations of historical stuff than the node's so we might also want to replace the node's scanner w this one.. but I like the responsiveness & tweakableness of the client-side version so I'm not 100% sure I want to get rid of it from the dashboard entirely.. this might be something else we could incorporate into the cf-core utils & import into both the node & dashboard.

modules/dashboard/src/components/Admin/Admin.js
bohendo: The scanner in this PR is more powerful than the nodes since it can check and confirm whether a channel is unrecoverable (eg if a toxic bytecode + mismatched factory was used to calculate a given multisig address).. I suspect that once we port it over to the node & re-run it we'll find that those 4 channels are unrecoverable.

bohendo: Note that this upgraded getCreate2Address will use a given (toxic) bytecode or default to fetching it from the factory if the given one is falsy (eg `""`)

PR #652, opened 2019-12-17
651 fix channel provider

### LINKED ISSUE #646: [cf-core, client] Use Hex strings instead of BigNumber types
## Description

Currently `cf-core` expects BigNumber types to be passed through rpcRouter however this is not JSON-RPC compatible which breaks how other ChannelProvider implementations would work which would require BigNumber types to be stringfied before transport.

Additionally the `client` should ensure that all BigNumber types passed as JSON-RPC parameters should be converted to hex strings.

## Problem

This issue is present in both Torus and WalletConnect integrations for example when executing a condtionalTransfer which expects `initiatorDeposit` to be BigNumber type to have a method `toHexString` inside CFCore.

## Solution

All instances where a BigNumber type is expected or is an output should be replaced as a hex string internally and externally hence they would use `bigNumerify` method from ethers when necessary.

### LINKED ISSUE #651: channelProvider is broken
To reproduce:

1. open two daicards (one will be your smart client daicard, and one will be your dumb client -- provider -- daicard)
2. in the dumb client card, go to settings and click "activate wallet connext"
3. take a picture with your phone of the qr code that pops up
4. in your smart client card, open up the scanner and scan the pic of the qr code

it should spin up the same smart client card address in both windows. you should go through the flows on the dumb client (but keep in mind that the dumb client card does not auto-deposit, so you'll have to do the deposit into the smart client card)

the issue i was seeing was happening in getAppInstances which is called on connect (specifically with reference to finding the any installed registry or coin balance refund apps)

so to repro my exact issue just wait, but its probably best to go through the flows

bohendo:
DO NOT MERGE UNTIL NEW WALLET CONNECT [PUBLISHED](https://github.com/WalletConnect/walletconnect-monorepo/pull/210#pullrequestreview-333773336)

closes #651 
closes #646 

## Checklist:


LayneHaber:
Also includes some cleanup from other issues that were opened/closed while the larger connect refactor branch was open

modules/client/src/controllers/WithdrawalController.ts
```diff
@@ -67,7 +66,8 @@ export class WithdrawalController extends AbstractController {
       }
       const postWithdrawBalances = await this.connext.getFreeBalance(assetId);
 
-      const expectedFreeBal = preWithdrawBalances[myFreeBalanceAddress].sub(amount);
+      console.log(`***** preWithdrawBalances: ${stringify(preWithdrawBalances)}`);
```
bohendo: We should probably prefer `this.log` over `console`

modules/client/src/node.ts
```diff
@@ -279,10 +279,13 @@ export class NodeApiClient implements INodeApiClient {
         `Must have instantiated a channel provider (ie a signing thing) before setting auth token`,
       );
     }
+    // TODO: merge:
+    // https://github.com/WalletConnect/walletconnect-monorepo/pull/210
+    // and publish / update package
     const nonce = await this.send("auth.getNonce", {
       address: this.channelProvider.signerAddress,
     });
```
bohendo: How is this TODO comment related to `auth.getNonce`?

bohendo: How is this TODO comment related to `auth.getNonce`?

modules/daicard/src/App.js
```diff
@@ -271,15 +271,15 @@ class App extends React.Component {
...
-      let rpc = {};
-      rpc[network.chainId] = urls.ethProviderUrl;
-      const channelProvider = new WalletConnectChannelProvider();
+      const rpc = {
+        [network.chainId]: urls.ethProviderUrl
+      }
+      const channelProvider = new WalletConnectChannelProvider({
+        rpc,
+        chainId: network.chainId,
+      });
```
bohendo: Why not use this rpc value directly in the one place it's needed. Looks like we aren't gonna need that `rpc` var again later so I vote we remove it.

pedrouid:
Published channel-providers updates on WalletConnect monorepo to NPM and tested it 👍 Good to merge

modules/client/src/controllers/WithdrawalController.ts
LayneHaber: oh we should definitely take that out that's not supposed to be in there 🙃

modules/client/src/node.ts
LayneHaber: The channel provider (when injected) would not allow access to `this.channelProvider.signerAddress`

LayneHaber: see above:

> The channel provider (when injected) would not allow access to this.channelProvider.signerAddress until the latest wc is published

PR #662, opened 2019-12-23
[integration-tests] Integration test runner

### LINKED ISSUE #655: Create integration test harness
Use the node's test framework for this. Features/Requirements:

- In-memory database
- In-memory messaging
- Ganache contained and configured within test runner
- Clients started up and torn down within test runner
- Ability to mock messaging providers during test runs

bohendo:
## The Problem

closes #655

## Checklist:


modules/integration-test/src/util/store.ts
```diff
@@ -0,0 +1,54 @@
+import { CFCoreTypes, StorePair } from "@connext/types";
+
+export class MemoryStoreService implements CFCoreTypes.IStoreService {
```
bohendo: Looks like this was copy/pasted from `modules/cf-core/test/services/memory-store-service.ts`, we should create a `@connext/store` package (same flavor as `@connext/messaging`) that includes one implementation of this in-memory-store so we don't have to maintain it in two places.

modules/node/package.json
```diff
@@ -44,6 +44,7 @@
     "uuid": "3.3.3"
   },
   "devDependencies": {
+    "@connext/client": "1.4.1",
```
bohendo: If we're doing integration tests in `modules/integration-tests` then we should prob keep `module/node` unit tests focused on the node units rather than node-client interactions. Seems like this import could be removed eventually

modules/node/src/test/deployContracts.ts
```diff
@@ -0,0 +1,184 @@
...
+const artifacts = {};
+for (const contract of coreContracts) {
+  try {
+    artifacts[contract] = require(`@connext/cf-adjudicator-contracts/build/${contract}.json`);
+    console.log(`Imported adjudicator contract: ${contract}`);
+  } catch (e) {
+    artifacts[contract] = require(`@connext/cf-funding-protocol-contracts/build/${contract}.json`);
+    console.log(`Imported funding contract: ${contract}`);
+  }
+}
```
bohendo: Fun fact: new work I've been doing in my CI/CD branch includes wrapping the ethprovider stuff in a docker image. That means no need to copy/paste this contract deployment stuff. Instead, we start up our ethprovider image (instead of default truffle/ganache:latest) & that will automatically deploy contracts on startup.

modules/node/src/test/integration/deposit/deposit.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+    const clientAOpts: ClientOptions = {
+      ethProviderUrl: config.getEthRpcUrl(),
+      logLevel: 4,
+      mnemonic:
+        "humble sense shrug young vehicle assault destroy cook property average silent travel",
+      nodeUrl,
+      store: clientAStore,
+    };
+    clientA = await connext.connect(clientAOpts);
+    await clientA.isAvailable();
```
bohendo: Have we decided whether integration tests will go here or in their own module? I don't have strong feelings either way other than suggesting we not do both

modules/node/src/test/store.ts
```diff
@@ -0,0 +1,54 @@
+import { CFCoreTypes, StorePair } from "@connext/types";
+
+export class MemoryStoreService implements CFCoreTypes.IStoreService {
```
bohendo: Oooh another copy/paste of the in-memory store. Here's a little extra incentive to DRY out the store code by moving it into it's own module ASAP

rhlsthrm: Yep, this one is deleted though. I'll do that though.

modules/node/src/test/integration/deposit/deposit.spec.ts
rhlsthrm: Deleted this, going with the "own module" route.

rhlsthrm:
> Once the new ethprovider image is merged in, we should use this instead of copy/pasting the contract deployment script
We aren't copy-pasting the contract deployment anymore, assuming this isn't a blocking issue for the merge.

PR #667, opened 2019-12-26
641 always resolve promises

ArjunBhuptani:
Resolve/reject promises in all cases - this fixes several instances in controllers where promises just hang if an error occurs.


modules/node/src/transfer/transfer.service.ts
```diff
@@ -216,10 +216,14 @@ export class TransferService {
           "DEPOSIT_CONFIRMED_EVENT",
           async (msg: DepositConfirmationMessage) => {
             if (msg.from !== this.cfCoreService.cfCore.publicIdentifier) {
-              return;
+              reject(
+                `Deposit event from field: ${msg.from}, did not match public identifier: ${this.cfCoreService.cfCore.publicIdentifier}`,
+              );
```
LayneHaber: we don't really want to reject here, since you could be getting a `DEPOSIT_CONFIRMED_EVENT` from collateralizing other channels, or in this case from the user of other channels

modules/node/src/transfer/transfer.service.ts
```diff
@@ -216,10 +216,14 @@ export class TransferService {
...
             }
             if (msg.data.multisigAddress !== channel.multisigAddress) {
-              return;
+              reject(
+                `Deposit event multisigAddress: ${msg.data.multisigAddress}, did not match channel multisig address: ${channel.multisigAddress}`,
+              );
```
LayneHaber: again, seems a little weird to reject here since it  could be an irrelevant event thrown.

theres no error handling for these rejections so the payment would just fail

modules/node/src/transfer/transfer.service.ts
```diff
@@ -228,8 +232,9 @@ export class TransferService {
               assetId,
             );
             if (fb[freeBalanceAddr].lt(amountBN)) {
-              // wait for resolve
-              return;
+              reject(
+                `Free balance associated with ${freeBalanceAddr} is less than transfer amount: ${amountBN}`,
+              );
```
LayneHaber: fine to reject here (imo) because we don't ever re-request a deposit into the channel so the payment would fail anyways.

we may want to handle that edge case better

Issue #668, opened 2019-12-26
Deposit Integration Tests (low hanging fruit)


ArjunBhuptani:
Do these:
- [x] Happy case: bot A proposes deposit with node and deposit completes successfully with Eth
- [x] Happy case: bot A proposes deposit with node and deposit completes successfully with tokens
- [x] Bot A proposes deposit with invalid token address
- [x] Bot A proposes deposit with negative value
- [x] Bot A proposes deposit with value it does not have
- [x] Bot A already has already requested deposit rights before calling deposit
- [x] Bot A deposits eth, withdraws, and then deposits eth again
- [x] Bot A deposits eth, withdraws, and then deposits tokens

PR #669, opened 2019-12-27
668 deposit tests

ArjunBhuptani:
https://github.com/ConnextProject/indra/issues/668 has a checklist

Update: Only doing low hanging deposit tests in this PR so we can merge more frequently. Will amend the issue above too.


bohendo:
Looks good, I'll keep an eye on it & merge once tests pass

PR #674, opened 2020-01-03
Replace eth-crypto with eccrypto

pedrouid:
Ensuring it works with React-Native depends on the following PR from `eccrypto`:
https://github.com/bitchan/eccrypto/pull/50

However can be added to a postinstall script as follows:
```sh
TARGET="node_modules/eccrypto/browser.js"
sed -i'' -e "s/typeof window === 'undefined'/typeof browserCrypto === 'undefined' || typeof browserCrypto.getRandomValues === 'undefined'/" $TARGET;
```


pedrouid:
replaced by #678

PR #678, opened 2020-01-06
677 replace eth crypto

### LINKED ISSUE #677: [client] eth-crypto incompatible with react-native
Currently eth-crypto has a WebCrypto API incompatibility with react-native. This could be solved by porting the specific methods and use the eccrypto dependency independently

To make sure it works with React-Native, this issue depends on the following PR from `eccrypto`:
https://github.com/bitchan/eccrypto/pull/50

However can be added to a postinstall script to apply the fix as follows:
```sh
TARGET="node_modules/eccrypto/browser.js"
sed -i'' -e "s/typeof window === 'undefined'/typeof browserCrypto === 'undefined' || typeof browserCrypto.getRandomValues === 'undefined'/" $TARGET;
```

pedrouid:
Fix #677


LayneHaber:
still need to test in react native though, couldn't get to it today

modules/client/src/eth-crypto.ts
```diff
@@ -0,0 +1,126 @@
...
+import { publicKeyConvert } from "secp256k1";
+
+type Encrypted = {
+  ciphertext: string;
+  ephemPublicKey: string;
+  iv: string;
+  mac: string;
+};
+
+export function removeTrailing0x(str: string): string {
```
bohendo: `Trailing0x`? Think you mean `Leading0x`

modules/client/src/eth-crypto.ts
```diff
@@ -0,0 +1,126 @@
...
+  if (str.startsWith("0x")) {
+    return str.substring(2);
+  }
+  return str;
+}
+
+export function addTrailing0x(str: string): string {
```
bohendo: This function shouldn't be exported since it's never used outside this module

modules/client/src/connext.ts
```diff
@@ -831,7 +831,7 @@ export class ConnextClient implements IConnextClient {
       throw new Error(`No way to decode transfer, this should never happen!`);
     }
 
-    const cipher = EthCrypto.cipher.parse(encryptedPreImage);
+    const cipher = EthCrypto.parse(encryptedPreImage);
 
     const preImage = await EthCrypto.decryptWithPrivateKey(privateKey, cipher);
```
bohendo: Why generate this intermediate `cipher` var? If it's never used, then we should hide it w/in the EthCrypto module & simplify the `decryptWithPrivateKey` function sig accordingly

modules/client/src/controllers/ConditionalTransferController.ts
```diff
@@ -1,7 +1,7 @@
-import EthCrypto from "eth-crypto";
 import { HashZero, Zero } from "ethers/constants";
 import { fromExtendedKey } from "ethers/utils/hdnode";
 
+import * as EthCrypto from "../eth-crypto";
```
bohendo: Which functions from the crypto module do we need? Specifying im/exports explicitly draws attention to the interface between this module and the rest of the codebase which encourages us to keep it as simple as possible which makes the code easier to maintain.

modules/client/src/controllers/ConditionalTransferController.ts
```diff
@@ -90,7 +90,7 @@ export class ConditionalTransferController extends AbstractController {
       recipientPublicKey.slice(2), // remove 0x
       preImage,
     );
-    const encryptedPreImage = EthCrypto.cipher.stringify(encryptedPreImageCipher);
+    const encryptedPreImage = EthCrypto.stringify(encryptedPreImageCipher);
     await this.connext.setRecipientAndEncryptedPreImageForLinkedTransfer(
       recipient,
       encryptedPreImage,
```
bohendo: The `encryptedPreImage` var isn't ever used here, it should be hidden w/in the crypto module

PR #679, opened 2020-01-06
609 export data

hthillman:
## Checklist:


rhlsthrm:
Are we planning to have dashboard cypress tests? That seems like overkill at the moment...

hthillman:
That was my take ^.  Until we're using it to make decisions about how to move money around, I don't think it needs a test suite.

PR #681, opened 2020-01-06
680 test review fixes

### LINKED ISSUE #680: test review fixes
**Describe the bug**
There were a few issues with the `resolveTransfer` logic found when doing code reviews. The client was delivering the unencrypted preimage to the node and trusting the node to install, take action, and uninstall the app.

Making sure the client does this instead of the hub is the primary change, but other simplifications of the code and small tweaks are also needed.

LayneHaber:
## The Problem
Fixes #680 

Code was messy and resolving links was trusted.

## The Solution
Client installs, takes action, and uninstalls the linked transfer app rather than relying on the hub to do it.

Also cleans up some of the existing code.

## Checklist:


modules/client/src/controllers/ResolveConditionController.ts
```diff
@@ -112,11 +181,89 @@ export class ResolveConditionController extends AbstractController {
...
+
+    if (paymentIdParam !== paymentId) {
+      throwErr(`incorrect paymentId`);
+    }
+
+    if (linkedHash !== createLinkedHash(amountParam, assetIdParam, paymentIdParam, preImageParam)) {
+      throwErr(`incorrect linked hash`);
+    }
+
+    if (mkHash("0x0") !== preImage) {
```
bohendo: Not a huge fan of this mkHash util, the `HashZero` ethers constant would be better here IMO

PR #691, opened 2020-01-07
686 async transfer tests happy case

rhlsthrm:
## Checklist:


modules/test-runner/src/transfer/asyncTransfer.test.ts
```diff
@@ -0,0 +1,72 @@
...
+      [nodeFreeBalanceAddress]: postTransferFreeBalanceEthNodeA,
+    } = await clientA.getFreeBalance(AddressZero);
+
+    const {
+      [clientB.freeBalanceAddress]: postTransferFreeBalanceEthClientB,
+      [nodeFreeBalanceAddress]: postTransferFreeBalanceEthNodeB,
+    } = await clientB.getFreeBalance(AddressZero);
+
+    expect(postTransferFreeBalanceEthClientA).toBeBigNumberEq(0);
+    expect(postTransferFreeBalanceEthNodeA).toBeBigNumberEq(0); // not transferAmount?
```
bohendo: Is the node-sender channel supposed to be empty after these tests? bc it is..

modules/client/src/node.spec.ts
```diff
@@ -7,13 +7,11 @@ describe("NodeApiClient", () => {
...
     nodeClient = new NodeApiClient({
       messaging: new MockMessagingService(),
-      nodeUrl,
-      publicIdentifier: "xpubsomething",
     });
     expect(nodeClient).toBeInstanceOf(NodeApiClient);
   });
 
-  test("should call the config method properly", async () => {
+  test.skip("should call the config method properly", async () => {
```
LayneHaber: if we skip tests we should make it a habit to leave comments for why its skipped

Issue #693, opened 2020-01-08
[test-runner] Low hanging swap cases


ArjunBhuptani:
- [x] Bot A tries to swap with invalid from token address
- [x] Bot A tries to swap with invalid to token address
- [x] Bot A tries to swap with insufficient free balance for user
- [x] Bot A tries to swap with insufficient collateral on node
- [x] Bot A tries to swap with negative swap rate
- [x] Bot A tries to swap with negative userAmount

PR #696, opened 2020-01-08
693 basic swap cases

ArjunBhuptani:
https://github.com/ConnextProject/indra/issues/693

Also includes skipped test for invalid swap rate. Right now, this always fails because client never resolve/rejects promise when node rejects install. Opened an issue for this here: https://github.com/ConnextProject/indra/issues/700


modules/test-runner/src/swap/swap.test.ts
```diff
@@ -130,4 +130,123 @@ describe("Swaps", () => {
     );
     expect(postSwapFreeBalanceTokenNode).toBeBigNumberEq(swapAmount);
   });
+
+  test("Bot A tries to swap with invalid from token address", async () => {
+    // client deposit and request node collateral
+    await clientA.deposit({ amount: parseEther("0.01").toString(), assetId: AddressZero });
+    await clientA.requestCollateral(tokenAddress);
+
+    const swapRate = await clientA.getLatestSwapRate(AddressZero, tokenAddress);
```
LayneHaber: we should be checking this swap rate is as expected

modules/test-runner/src/swap/swap.test.ts
```diff
@@ -130,4 +130,123 @@ describe("Swaps", () => {
...
+    const wrongAddress = "0xdeadbeef";
+    const swapAmount = One;
```
LayneHaber: really? one wei?

modules/test-runner/src/swap/swap.test.ts
```diff
@@ -130,4 +130,123 @@ describe("Swaps", () => {
...
+  });
+
+  test("Bot A tries to swap with invalid to token address", async () => {
+    // client deposit and request node collateral
+    await clientA.deposit({ amount: parseEther("0.01").toString(), assetId: AddressZero });
+    await clientA.requestCollateral(tokenAddress);
+
+    const swapRate = await clientA.getLatestSwapRate(AddressZero, tokenAddress);
+    const wrongAddress = "0xdeadbeef";
+    const swapAmount = One;
```
LayneHaber: we should make the balances more realistic

modules/test-runner/src/swap/swap.test.ts
```diff
@@ -130,4 +130,123 @@ describe("Swaps", () => {
...
+      swapRate,
+      toAssetId: wrongAddress,
+    };
+    await expect(clientA.swap(swapParams)).rejects.toThrowError(`is not a valid eth address`);
+  });
+
+  test("Bot A tries to swap with insufficient free balance for the user", async () => {
+    // client deposit and request node collateral
+    await clientA.deposit({ amount: parseEther("0.01").toString(), assetId: AddressZero });
+    await clientA.requestCollateral(tokenAddress);
```
LayneHaber: it feels like this could be done in `before` setup

modules/test-runner/src/swap/swap.test.ts
```diff
@@ -130,4 +130,123 @@ describe("Swaps", () => {
...
+
+    const swapRate = await clientA.getLatestSwapRate(AddressZero, tokenAddress);
+    const swapAmount = One;
+    const swapParams: SwapParameters = {
+      amount: swapAmount.toString(),
+      fromAssetId: AddressZero,
+      swapRate: (-swapRate).toString(),
```
LayneHaber: does this even work?

LayneHaber: there should be a `BigNumber` function for it that we should definitely use

modules/test-runner/src/swap/swap.test.ts
ArjunBhuptani: How exactly can we check swap rate is as expected? This fn pulls it from node lol

modules/test-runner/src/swap/swap.test.ts
ArjunBhuptani: Why? It shouldn't make a difference, and the one wei helps us catch potential rounding errors

modules/test-runner/src/swap/swap.test.ts
ArjunBhuptani: It can't because not every test deposits eth and not every test requests collateral.

modules/test-runner/src/swap/swap.test.ts
ArjunBhuptani: I can change it for the some of the other tests to add some variety

modules/test-runner/src/swap/swap.test.ts
ArjunBhuptani: It does work. Swap rate is a string not `BigNumber`
Screenshot from 2020-01-10 13-04-35

modules/test-runner/src/swap/swap.test.ts
ArjunBhuptani: It really shouldn't make any difference here

ArjunBhuptani: I'd even argue that we should be testing smaller amounts to catch potential rounding issues

Issue #703, opened 2020-01-09
[test-runner] ChannelProvider tests


ArjunBhuptani:
- [x] Happy case: Bot A1 can be instantiated with a channelProvider generated from Bot A
- [x] Happy case: Bot A1 can call the full deposit → swap → transfer → withdraw flow on Bot A
- [x] Bot A1 tries to call a function when Bot A is offline
- [ ] Bot A1 tries to reject installing a proposed app that bot A has already installed?

LayneHaber:
Skipping this test for now and closing issue

``` Bot A1 tries to reject installing a proposed app that bot A has already installed```

PR #708, opened 2020-01-10
703 channel provider tests

pedrouid:
## Checklist:


pedrouid:
Fix #703

modules/client/src/channelProvider.ts
```diff
@@ -40,33 +45,77 @@ export const createCFChannelProvider = async ({
...
+  };
+
+  public once = (event: string, listener: (...args: any[]) => void): any => {
+    this.cfCore.once(event as any, listener);
+    return this.cfCore;
+  };
+
+  public open(): void {
+    this.connected = true;
+  }
```
LayneHaber: do we have to explicitly call `.open` when using the `RpcConnection` ? If this is all it does it seems a little unneccessary, but i get it if it's part of adhering to an accepted interface

modules/client/src/channelProvider.ts
```diff
@@ -134,12 +187,12 @@ export class ChannelProvider {
...
   ///////////////////////////////////////////////
   ///// LISTENER METHODS
-  public on = (event: string, listener: (...args: any[]) => void): RpcConnection => {
+  public on = (event: string, listener: (...args: any[]) => void): any => {
     this.connection.on(event, listener);
     return this.connection;
   };
 
-  public once = (event: string, listener: (...args: any[]) => void): RpcConnection => {
+  public once = (event: string, listener: (...args: any[]) => void): any => {
```
LayneHaber: we can do a better job of the `event` and `return` types here as well probably

modules/test-runner/src/channelProvider/channelProvider.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    ////////////////////////////////////////
+    // WITHDRAW FLOW
+    const withdraw = { amount: bigNumberify(ONE), assetId: AddressZero };
+    await withdrawFromChannel(clientA1, withdraw.amount.toString(), withdraw.assetId);
+  });
+
+  // tslint:disable-next-line:max-line-length
+  test("Bot A1 tries to call a function when Bot A is offline", async () => {
+    // close channelProvider connection
+    clientA1.channelProvider.close();
```
LayneHaber: ahhhhh so the `open` / `close` is to simulate going "offline"?

modules/test-runner/src/swap/swap.test.ts
```diff
@@ -18,130 +22,41 @@ describe("Swaps", () => {
...
-      [nodeFreeBalanceAddress]: preSwapFreeBalanceTokenNode,
-    } = await clientA.getFreeBalance(tokenAddress);
-    expect(preSwapFreeBalanceTokenNode).toBeBigNumberEq(Zero);
-    expect(preSwapFreeBalanceTokenClient).toBeBigNumberEq(parseEther("10"));
-
-    const swapRate = await clientA.getLatestSwapRate(AddressZero, tokenAddress);
-    const inverseSwapRate = inverse(swapRate);
-    console.log("inverseSwapRate: ", inverseSwapRate);
+    await clientA.deposit({ amount: input.amount.toString(), assetId: input.assetId });
+    await clientA.requestCollateral(output.assetId);
```
LayneHaber: great cleanup! 🥇

modules/test-runner/src/transfer/asyncTransfer.test.ts
```diff
@@ -22,170 +21,20 @@ describe("Async Transfers", () => {
...
-      assetId: tokenAddress,
-      amount: transferAmount.toString(),
-      status: "RECLAIMED",
-      type: "LINKED",
-      senderPublicIdentifier: clientA.publicIdentifier,
-      receiverPublicIdentifier: clientB.publicIdentifier,
-      meta: { hello: "world" },
-    });
-    expect(paymentB).toMatchObject(paymentA);
+    await asyncTransferAsset(clientA, clientB, transferAmount, assetId, nodeFreeBalanceAddress);
```
LayneHaber: also great cleanup 🥇

modules/test-runner/src/util/bn.ts
```diff
@@ -3,3 +3,7 @@ import { BigNumber, bigNumberify, BigNumberish, formatEther, parseEther } from "
 export const toWei = (n: BigNumberish): BigNumber => parseEther(n.toString());
 
 export const inverse = (bn: BigNumberish): string => formatEther(toWei(toWei("1")).div(toWei(bn)));
+
+export const calculateExchange = (amount: BigNumber, swapRate: string): BigNumber => {
+  return bigNumberify(formatEther(amount.mul(parseEther(swapRate))).replace(/\.[0-9]*$/, ""));
+};
```
LayneHaber: isn't this a function exported from the client? seems like we should use that (but don't have a huge pref here, i think we just have the same fn in three places: here, client, daicard)

modules/test-runner/src/util/channelProvider.ts
```diff
@@ -0,0 +1,197 @@
...
+  public open(): void {
+    this.connected = true;
+  }
+
+  public close(): void {
+    this.connected = false;
+  }
+}
+
+export class ChannelProvider extends EventEmitter implements IChannelProvider {
```
LayneHaber: can we name this `MockChannelProvider` or something if it's only going to be used for tests?

modules/test-runner/src/util/client.ts
```diff
@@ -45,6 +48,21 @@ export const createClient = async (opts?: Partial<ClientOptions>): Promise<IConn
...
+  };
+
+  const client = await connect(clientOpts);
+
+  await client.isAvailable();
+
+  expect(client.freeBalanceAddress).toBeTruthy();
+  expect(client.publicIdentifier).toBeTruthy();
+
+  return client;
```
LayneHaber: niceee

modules/types/src/channelProvider.ts
```diff
@@ -43,39 +84,28 @@ export const ChannelProviderRpcMethods = {
 };
 export type ChannelProviderRpcMethod = ConnextRpcMethod | CFCoreTypes.RpcMethodName;
 
+export type JsonRpcRequest = {
+  id: number;
+  jsonrpc: "2.0";
+  method: string;
+  params: any;
+};
```
LayneHaber: is this exported from cf or no?

modules/client/src/channelProvider.ts
```diff
@@ -40,33 +45,77 @@ export const createCFChannelProvider = async ({
...
+  public cfCore: CFCore;
+
+  constructor(cfCore: CFCore) {
+    super();
+    this.cfCore = cfCore;
+  }
+
+  public async send(payload: JsonRpcRequest): Promise<any> {
+    if (!this.connected) {
+      return Promise.resolve();
```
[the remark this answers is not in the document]
pedrouid: You're right

modules/client/src/channelProvider.ts
pedrouid: Yeah it's more for matching the interface, I can also make it a no-op for CF channelProvider

modules/client/src/channelProvider.ts
pedrouid: Actually this could just be void

pedrouid: Actually it can't because it extends from EventEmitter

modules/test-runner/src/channelProvider/channelProvider.test.ts
pedrouid: yes!

modules/types/src/channelProvider.ts
pedrouid: vscode wasn't returning anything but I can search again

pedrouid: Actually it doesn't because it uses a different schema
```typescript
export type Rpc = {
  methodName: string;
  parameters: RpcParameters;
  id?: number;
};
```

modules/test-runner/src/util/bn.ts
pedrouid: not sure here just used what was available before in the `swap.test.ts` cc @rhlsthrm

Issue #707, opened 2020-01-10
[test-runner] Get State Channel tests


ArjunBhuptani:
- [x] Happy case: should return `stateChannelJSON` from store with multisig address
- [x] Store does not contain state channel
- [x] Store contains state channel on wrong multisig address
- [x] Store contains multiple state channels
- [x] State channel under multisig key has no proxy factory address
- [x] State channel under multisig key has no freeBalanceAppInstance

PR #709, opened 2020-01-11
707 get state channel tests

ArjunBhuptani:
Closes https://github.com/ConnextProject/indra/issues/707

One issue I ran into was that for:
1) State channel under multisig key has no proxy factory address
2) State channel under multisig key has no freeBalanceAppInstance
3) Store contains state channel on wrong multisig address
The promise just resolves with the incorrect info so tests are not particularly useful.

Either we should build validators/recovery for the above cases where channels break, or just delete these tests in the future. For now, they're written but skipped.

Tracking issue for the above task is here: https://github.com/ConnextProject/indra/issues/710


modules/test-runner/src/util/client.ts
```diff
@@ -13,19 +13,18 @@ const wallet = Wallet.fromMnemonic(env.mnemonic).connect(ethProvider);
 
 let clientStore: MemoryStoreService;
 
-export const createClient = async (
-  mnemonic: string = Wallet.createRandom().mnemonic,
-): Promise<IConnextClient> => {
+export const createClient = async (opts?: ClientOptions): Promise<IConnextClient> => {
```
LayneHaber: shouldn't the type be something like: `opts?: Partial<ClientOptions>` so you can provide a subset of opts to start the client with that would override any defaults?

technically that's happening when setting `clientOpts`, but the typing seems off

PR #718, opened 2020-01-12
[daicard, node, client] 571 collateralization issues

### LINKED ISSUE #571: Daicard seems to recollateralize multiple times when it's doing the first time swap after deposit
**Describe the bug**
Seeing multiple collateralization logs. See attached screenshot.

**To Reproduce**
1. Deposit in fresh Daicard channel
2. Watch logs

**Expected behavior**
Only collateralize once.

**Screenshots**
image

LayneHaber:
## The Problem

Fixes #571 

Additionally, found a few other things:
- `addPaymentProfile` was in the client API, meaning users could arbitrarily add their own profiles that would be honored when making a `requestCollateral` call (should be an admin function)
- `autoswap` was being called twice, one right after the other at the end of `autoDeposit` and at the next line in the poller
- `autoswap` was also not able to handle edge case where it has insufficient collateral to swap their entire eth balance (but will not add additional collateral following their profile)

## The Solution

- 🔴  BREAKING CHANGE!! 🔴  `addPaymentProfile` removed from client API and modified to use admin auth 
- `autoswap` function cleaned up

## Checklist:

### DIFF
--- a/modules/client/src/connext.ts
+++ b/modules/client/src/connext.ts
@@ -259,10 +259,6 @@ export class ConnextClient implements IConnextClient {
     return await this.node.unsubscribeFromSwapRates(from, to);
   };
 
-  public addPaymentProfile = async (profile: PaymentProfile): Promise<PaymentProfile> => {
-    return await this.node.addPaymentProfile(profile);
-  };
-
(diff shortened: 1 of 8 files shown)


bohendo:
Note that backwards compatibility checks are being enforced, see PR #694 description for more info.

PR #732, opened 2020-01-14
724 store module

### LINKED ISSUE #724: [store] Add store module
Add store package source code from [connext-store](https://github.com/pedrouid/connext-store) into a monorepo module

pedrouid:
Fix #724


rhlsthrm:
@pedrouid it looks like the stores are all async now, does that mean we need to `await` all the `get`s in the CF code?

PR #733, opened 2020-01-14
715 async transfer easy cases

### LINKED ISSUE #715: [test-runner] Async transfer low-hanging fruit tests
- [x] Bot A tries to transfer a negative amount
- [x] Bot A tries to transfer with an invalid token address
- [ ] Bot A tries to transfer with an unsupported (but not invalid) token address
- [x] Bot A tries to transfer with invalid recipient xpub
- [x] Bot A tries to transfer an amount greater than they have in their free balance
- [x] Bot A tries to transfer with a payment Id that is not 32 bytes
- [x] Bot A tries to transfer with a preimage that is not 32 bytes
- [x] Bot A proposes a transfer to an xpub that doesn’t have a channel

LayneHaber:
Fixes #715 

Skipping a test, issue opened #741


modules/client/src/controllers/AbstractController.ts
```diff
@@ -45,7 +45,7 @@ export abstract class AbstractController {
             `indra.node.${this.connext.nodePublicIdentifier}.install.${appInstanceId}`,
             boundResolve,
           );
-          this.listener.on(CFCoreTypes.EventNames.REJECT_INSTALL_EVENT, boundReject);
+          this.listener.on("REJECT_INSTALL_EVENT", boundReject);
```
pedrouid: Is this a typo or intended?

modules/test-runner/src/transfer/asyncTransfer.test.ts
```diff
@@ -27,30 +34,141 @@ describe("Async Transfers", () => {
   }, 90_000);
 
   test("happy case: client A transfers eth to client B through node", async () => {
-    const transfer: AssetOptions = { amount: ETH_AMOUNT_SM, assetId: AddressZero };
```
pedrouid: can we add this back just to make sure the amount and assetId is this the same throughout the test

LayneHaber: just the `transfer` object? we can if you think it makes it easier to read, i personally prefer it the other way

keeping the `fundChannel` though because it checks to make sure money is deposited by the client correctly

modules/client/src/controllers/AbstractController.ts
LayneHaber: Intended, but can change it back

modules/test-runner/src/transfer/asyncTransfer.test.ts
pedrouid: I find it easier to read but mostly it makes sure the `amount` and `assetId` is not hardcoded in multiple places.

Yeah we should use the `fundChannel` in other tests too.

PR #736, opened 2020-01-15
Speedupify tests

bohendo:
## The Problem

Wow our tests were getting to be sooo sloooow

## The Solution

Apparently Jest has a huge performance overhead or something idk but using Mocha instead makes tests run >5x faster

## Checklist:


LayneHaber:
should fix #741

PR #739, opened 2020-01-15
one-liner default options

pedrouid:
Allowing clients to be instantiated with default options using our hosted hub by passing a network string as either `rinkeby` or `mainnet` (using mnemonic)

**Rinkeby**
```javascript
import * as connext from "@connext/client"

const channel = await connext.connect("rinkeby")
```

**Mainnet**
```javascript
import * as connext from "@connext/client"

const channel = await connext.connect("mainnet", { mnemonic })
```


pedrouid:
PS - Don't merge this PR without consolidating with #738

modules/client/src/default.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export async function getDefaultMnemonic(store: ConnextStore): Promise<string> {
+  const MNEMONIC_KEY = "MNEMONIC";
+  let mnemonic = await store.get(MNEMONIC_KEY);
+  if (!mnemonic) {
+    mnemonic = generateMnemonic();
+    store.set([{ path: MNEMONIC_KEY, value: mnemonic }]);
+  }
+  return mnemonic;
+}
```
LayneHaber: seems strange to have default options that aren't secure

pedrouid: Perhaps default is not right word here, this is more like a beginner options for hackathons and demos, realistically we would advise production apps to manage their own mnemonics

However you can still pass override options while maintain the network string, here is an example for a react-native wallet
```javascript
const channel = await connext.connect('mainnet', {
  xpub,
  keygen,
 asyncStorage
})
```
Obviously the network string is not intended for production use but it can still be used to abstract the nodeUrl and ethProviderUrl to make it easier for us to bootstrap integrations without many breaking changes for our users

LayneHaber: yeah, we should make sure we add tests for all of this stuff.

i also think we should disable the use of default mnemonics for mainnet

pedrouid: Pushed these changes 👍

pedrouid:
Blocked by #742 to support testing

pedrouid:
Requires #754 and it should have tests of itself too

modules/test-runner/src/createChannel/createChannel.test.ts
```diff
@@ -10,6 +11,16 @@ import {
 } from "../util";
 
 describe("Create Channel", () => {
+  it("Happy case: user creates channel with mainnet network string", async () => {
+    const clientA: IConnextClient = await createDefaultClient("mainnet");
+    expect(clientA.multisigAddress).to.be.ok;
```
LayneHaber: we should also add a test that if you instantiate the default client and a client with client opts you get the same multisig address

PR #744, opened 2020-01-16
restructure store module

pedrouid:
Can I get this one merged before I finish 742? Just to remove the noise. This is just moving files around.

PS - also abstracted PisaClient as a BackupServiceAPI (as suggested by @bohendo)


modules/store/src/connextStore.ts
```diff
@@ -1,22 +1,22 @@
...
   DEFAULT_STORE_SEPARATOR,
-  PATH_CHANNEL,
-  PATH_PROPOSED_APP_INSTANCE_ID,
-} from "./constants";
-import InternalStore from "./store";
-import { IPisaClient, StoreFactoryOptions, StorePair } from "./types";
-import {
-  arrayify,
   hexlify,
+  IPisaClient,
```
bohendo: We should try to abstract out a "ThirdPartyBackupServiceAPI" type instead of including any Pisa-specific types or interfaces in our code

pedrouid: I like that 👍 Fortunately nobody is using this module with PisaClient yet so I can break backwards compatibility

PR #759, opened 2020-01-20
648 add mastercopy to state

### LINKED ISSUE #248: Remove TSLint
https://github.com/typescript-eslint/typescript-eslint#what-about-tslint

tl;dr TSLint is getting deprecated, and we should move to `typescript-eslint`

### LINKED ISSUE #624: [types, cf-core] Messages in types module
**Describe the bug**
All message types (i.e. `WithdrawFailedMessage`) emitted from events should live in the types module over the cf-types or cf-core module.

### LINKED ISSUE #648: Add multisig address to state
**Is your feature request related to a problem? Please describe.**

The minimum viable multisig code master copy is going to need an upgrade soonish (eg to address audit issues).

Issue is, the channel's multisig address is calculated based on the master copy's address. If we just deploy a new multisig master copy & start using it, all existing channels will break. No bueno.

**Describe the solution you'd like**

Give the multisig master copy the same treatment we gave the factory address: add it to the channel state.

Goal: if the hub crashes or gets banned in some country or whatev & users only have access to their local copy of the state, they still have all the info they need to withdraw. The multisig master copy address is just as critical to this process as the factory address.

**Describe alternatives you've considered**
We haven't really considered this problem yet.

bohendo:
closes #648 
closes #248 (kind of, it adds support for eslint but hasn't removed tslint yet)
closes #624

## The Problem

If we address audit issues & update the multisig mastercopy contract, all existing channels will break when we update the node's address book.

## The Solution

Add the multisig mastercopy address to the channel state (just like we did w the proxyFactory) so that all channels, even old ones using legacy schemes, can always withdraw their funds.

## Checklist:


modules/cf-core/src/models/state-channel.ts
```diff
@@ -369,12 +370,12 @@ export class StateChannel {
...
+    addresses: CriticalStateChannelAddresses,
     userNeuteredExtendedKeys: string[]
   ) {
     return new StateChannel(
       multisigAddress,
-      proxyFactoryAddress,
+      addresses,
       userNeuteredExtendedKeys,
       new Map<string, AppInstanceProposal>([]),
       new Map<string, AppInstance>(),
```
LayneHaber: If possible we should add a check here that the supplied multisig address is the expected one based on the critical state channel addresses

modules/cf-core/src/utils.ts
```diff
@@ -36,19 +36,20 @@ export function timeout(ms: number) {
...
  * @param {string} provider - to fetch proxyBytecode from the proxyFactoryAddress
  *
  * @returns {string} the address of the multisig
  *
  * NOTE: if the encoding of the multisig owners is changed YOU WILL break all
  * existing channels
  */
-// TODO: memoize
+// TODO: memoize?
+// TODO: replace addresses w CriticalStateChannelAddresses object?
```
LayneHaber: we should just replace now

modules/cf-core/test/integration/setup.ts
```diff
@@ -49,7 +49,7 @@ export async function setup(
   nodeCPresent: boolean = false,
   newExtendedPrvKey: boolean = false,
   messagingService: CFCoreTypes.IMessagingService = new MemoryMessagingService(),
-  storeServiceFactory: CFCoreTypes.ServiceFactory = new MemoryStoreServiceFactory(),
+  storeServiceFactory: any = new MemoryStoreServiceFactory(),
```
LayneHaber: does this have to be `any` casted?

modules/cf-core/src/models/state-channel.ts
bohendo: I'm confused, check them against what? We're being given 2 critical addresses to add to the state (via the `addresses` param), what do we compare these addresses to? Every time we call createEmptyState channel, we're passing in addresses from the NetworkContext so comparing them against that seems redundant.

modules/cf-core/test/integration/setup.ts
bohendo: Def not, good catch

modules/cf-core/src/models/state-channel.ts
LayneHaber: the `multisig` is dependent on those critical addresses right?

you can easily pass in a `multisig` that is different from what would be generated, and i would rather just know it would fail in that case

now it would probably have to be async so that may be a pain, and if were only using this within the package then its probably fine

PR #764, opened 2020-01-21
add backupService to ClientOptions

pedrouid:
## Checklist:

### DIFF
--- a/modules/client/src/connect.ts
+++ b/modules/client/src/connect.ts
@@ -85,6 +85,7 @@ export const connect = async (
       : clientOptions;
   const {
     asyncStorage,
+    backupService,
     logLevel,
     ethProviderUrl,
     nodeUrl,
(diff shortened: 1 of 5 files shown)


LayneHaber:
We should implement the default options in the daicard instead of [this](https://github.com/ConnextProject/indra/blob/07e7c170ec8c54dac90c6e152ffd0d4bbf805e10/modules/daicard/src/App.js#L251-L258)

since it is the reference implementation,

also before i merge, are the docs updated for the latest quick start process cc @pedrouid ?

pedrouid:
There is not much point implementing the default options for the daicard since they need to be overridden for local and staging environments

Not yet, I will update it now

LayneHaber:
coo ill go ahead and merge after this last ci/cd run goes through.

want to include #759 and #763  then cut a release

PR #771, opened 2020-01-22
fix strings constants + add channel-provider module

pedrouid:
Fix String Constants
* Replace strings with constants
* Fix keyof with actual values
* Stronger typing (no generic type casting)

Add Channel-Provider Module
* add channel-provider module
* replace channelprovider in client and test-runner
* move CFCore logic to RpcConnection


pedrouid:
> Can you update this section of the docs to reflect this?

I will after we cut a release

Issue #730, opened 2020-01-14
[channel-provider] remote event listeners


pedrouid:
Currently event listeners are implemented as no-ops for dumb clients like WalletConnect and Torus but as demonstrated by the test-runner channelProvider test these are necessary in dumb clients.

This can be handled by integrating a nats client to listen for events targeted at the client's nodePublicIdentifier.

See channelProvider.test.ts as example

pedrouid:
This should also refactor the logic handled by `setupMultisigAddress()` method in `connect.ts` to be handled by the remote event listener to await `CREATE_CHANNEL_EVENT` to get multisigAddress instead of being assigned externally

pedrouid:
Using a natsConnection similar to channelProvider.test.ts implementation doesn't cover all events emitted by CFCore.
Instead we should implement remove event listeners through RPC. Introducing a new JSON-RPC method called `chan_event_subscription` which has an object parameter with an event field of type CFCoreTypes.EventName.
Event subscriptions receive a response with equal payload id. The RemoveEventController will index event names  with subscribed payloadId then iterate through them to emit them separately as JSON-RPC responses.
The peer RPC connection will iterate through all the callbacks to trigger them consecutively

PR #773, opened 2020-01-22
add channel-provider module

pedrouid:
This module is a refactor as preparation for #730. It reduces a lot of code duplication on the ChannelProvider class and can be used by other providers to instantiate their own ChannelProviders without major breaking changes. Plus it will be easier to document.

Requires both #771 and #772 to be merged first


modules/channel-provider/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "clean": "rm -rf ./dist",
+    "prepare": "npm run build",
+    "test": "./node_modules/.bin/jest"
+  },
+  "dependencies": {
+    "@connext/types": "3.0.1"
+  },
+  "devDependencies": {
+    "@babel/polyfill": "7.7.0",
+    "@types/jest": "24.0.25",
```
LayneHaber: should probably just use `mocha` off the bat

modules/channel-provider/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "@types/node": "13.1.7",
+    "jest": "24.9.0",
+    "ts-jest": "24.2.0",
```
LayneHaber: same as above

PR #780, opened 2020-01-24
Update listeners.ts

balasan:
## The Problem
This was throwing an error in NodeJS (can't be parsing JSON objects)
Not sure if the check is necessary there... can it ever be a string?

## The Solution
removed JSON.parse but added a type check just in case
maybe there is a cleaner way?

## Checklist:


rhlsthrm:
Thanks for the PR! I was able to repro the issue and the fix you submitted didn't quite work, but it gave me a lead to make the proper fix: https://github.com/ConnextProject/indra/pull/783

PR #790, opened 2020-01-28
Expose removeListener in client api

balasan:
## The Problem
I find myself need to add temporary listeners and then remove them and it would be cleaner not to have to access then inner `listener` object

## The Solution
expose removeListener api

## Checklist:


LayneHaber:
looks good to me!

PR #805, opened 2020-01-30
688 smol swap

### LINKED ISSUE #688: Swap breaks with very small numbers with token -> ETH
Node is using float divisions to calculate rates:

```ts
const derivedRate =
      parseFloat(formatEther(responderDeposit)) / parseFloat(formatEther(initiatorDeposit));
```

This floors to 0 if numbers and swap rates are very small.

LayneHaber:
## The Problem
Fixes #688 

Also -- client seems to have a rounding error. From discord:
```
Error: Value (119329826098102879) is not less than or equal to 119250000000000000
    at Object.validate (validator.js:11)
    at SwapController._callee2$ (SwapController.js:72)
    at tryCatch (runtime.js:45)
    at Generator.invoke [as _invoke] (runtime.js:271)
    at Generator.prototype.<computed> [as next] (runtime.js:97)
    at fulfilled (SwapController.js:4)
    at es.promise.js:114
    at flush (microtask.js:26)
    at tryCallOne (core.js:37)
    at core.js:123
```

## The Solution
Make node calculate exchange same way client does, fix rounding error.

## Checklist:


modules/client/src/lib/index.ts
```diff
@@ -1,4 +1,3 @@
-export * from "./bn";
```
pedrouid: Is this necessary to be removed? It will break Metamask integration

rhlsthrm: Added back

Issue #792, opened 2020-01-29
[node] Onchain transactions


LayneHaber:
Make sure all onchain transactions are properly tracked and added to the database table in the node

PR #813, opened 2020-02-03
[node] 792 onchain transactions

rhlsthrm:
## The Problem
#792 

## The Solution
Make sure onchain transactions of all types (collateral, deposit, withdrawal) are stored in node's DB.

## Checklist:


modules/cf-core/src/methods/app-instance/uninstall/controller.ts
```diff
@@ -20,31 +20,30 @@ export default class UninstallController extends NodeController {
...
     requestHandler: RequestHandler,
-    params: CFCoreTypes.UninstallVirtualParams
+    params: CFCoreTypes.UninstallVirtualParams,
   ): Promise<string[]> {
     const { store } = requestHandler;
     const { appInstanceId } = params;
 
     const sc = await store.getChannelFromAppInstanceID(appInstanceId);
 
-    if (sc.freeBalance.identityHash === appInstanceId) {
```
LayneHaber: why did this get removed? still seems valid

modules/cf-core/src/methods/app-instance/uninstall/controller.ts
```diff
@@ -20,31 +20,30 @@ export default class UninstallController extends NodeController {
...
     if (!appInstanceId) {
       throw Error(NO_APP_INSTANCE_ID_TO_UNINSTALL);
     }
 
+    const sc = await store.getChannelFromAppInstanceID(appInstanceId);
+
+    if (sc.freeBalance.identityHash === appInstanceId) {
+      throw Error(CANNOT_UNINSTALL_FREE_BALANCE(sc.multisigAddress));
+    }
+
```
LayneHaber: jk i see it was just moved, which makes more sense

modules/node/src/cfCore/cfCore.service.ts
```diff
@@ -154,9 +146,40 @@ export class CFCoreService {
...
+    logger.debug(
+      `Calling ${ProtocolTypes.chan_withdraw} with params: ${stringify({
+        amount,
+        multisigAddress,
+        tokenAddress: assetId,
+      })}`,
+    );
+    const withdrawRes = await this.cfCore.rpcRouter.dispatch({
+      id: Date.now(),
+      methodName: ProtocolTypes.chan_deposit,
```
LayneHaber: shouldn't it be `chan_withdraw`

modules/node/src/channel/channel.service.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+      appDefinitionAddress: mkAddress("0xa"),
+      chainId: 1337,
+      name: "SimpleLinkedTransferApp",
+      outcomeType: OutcomeType.SINGLE_ASSET_TWO_PARTY_COIN_TRANSFER,
+      stateEncoding: "",
+    };
+  }
+}
+
+describe.skip("Channel Service", () => {
```
LayneHaber: this should be unskipped before merging

modules/node/src/channel/channel.service.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+        { provide: getRepositoryToken(OnchainTransaction), useClass: MockOnchainTransactionRepository },
+      ],
+    }).compile();
+    channelService = moduleRef.get<ChannelService>(ChannelService);
+    console.log("channelService: ", channelService);
+  });
+
+  it("should add deposits to the onchain transaction table", async () => {
+    const result = await channelService.deposit(mkAddress("0xAAA"), One);
+  });
```
LayneHaber: could probably use some more `expect`s here

modules/test-runner/src/util/helpers/requestDepositRights.ts
```diff
@@ -64,6 +62,7 @@ export const requestDepositRights = async (
       responderDepositTokenAddress: assetId,
       timeout: Zero,
     };
+    console.log("HEREEEEEEE");
```
LayneHaber: lol we can probably remove this

modules/node/src/cfCore/cfCore.service.ts
rhlsthrm: Clearly this has not been tested.

modules/node/src/channel/channel.service.spec.ts
rhlsthrm: This isn't a real test, I was playing around with the test framework a bit.

modules/node/src/channel/channel.service.spec.ts
rhlsthrm: Not a real test, the real test is in the integration tests.

PR #814, opened 2020-02-04
812 stale connection

### LINKED ISSUE #812: [test-runner] NatsError: Stale Connection
**Describe the bug**
[This](https://github.com/ConnextProject/indra/blob/8b675de2682ed610549d1a61be2eb03aa8b2633a/modules/test-runner/src/transfer/asyncTransferOffline.test.ts#L299) test is failing when running in the full suite with a `NatsError: Stale Connection` but passes individually.

LayneHaber:
## The Problem
Fixes #812 

## The Solution
Still a WIP!

## Checklist:


modules/messaging/src/nats.ts
```diff
@@ -22,7 +22,8 @@ export class NatsMessagingService implements IMessagingService {
     const config = this.config as nats.NatsConnectionOptions;
     config.servers = typeof messagingUrl === `string` ? [messagingUrl] : messagingUrl;
     config.payload = nats.Payload.JSON;
-    this.connection = await nats.connect(config);
+    // NOTE: high maxPingOut (default=2) to prevent errors while time-travelling during tests
+    this.connection = await nats.connect({ ...config, maxPingOut: 1000000 });
```
rhlsthrm: Hmm if this is only for tests, we should probably make it configurable at the client instantiation level and only override it in the tests. Also might want to look at `pingInterval` to make that super high so it doesn't even try to ping in the tests.

LayneHaber: According to [this](https://github.com/nats-io/nats-pure.rb/issues/14) issue the setting we really want is the `maxReconnectAttempts: -1`

But I ran it without modifying the config and it seems that the real issue was changing the amount of time the tests fast-forward (reducing it from 5mins)

Also see [this](https://github.com/nats-io/nats.rb/issues/143) issue

PR #815, opened 2020-02-05
use secp256k1 library for signing

pedrouid:
## Checklist:


modules/client/src/lib/utils.ts
```diff
@@ -86,3 +86,19 @@ export const createRandom32ByteHexString = (): string => {
...
 export const createPaymentId = createRandom32ByteHexString;
 export const createPreImage = createRandom32ByteHexString;
+
+export function removeTrailing0x(str: string) {
+  if (str.startsWith("0x")) {
+    return str.substring(2);
+  } else {
+    return str;
+  }
+}
```
bohendo: 1. This should be called remove**Leading**0x
2. This prob doesn't need to be a separate function bc we can just do `.replace(/^0x/, '')`

LayneHaber: Good catch -- completely missed this. I don't mind having an extra function for it, but we should update the name

pedrouid: Makes sense, I've renamed and refactored these hex utils in commit https://github.com/ConnextProject/indra/pull/815/commits/2cd997f308261f25d2912b592d19a2b958cd6169

pedrouid:
This was breaking with ConnextBenchmarking and even simply loading the DaiCard, I'm unsure how this even passed all the tests. We should investigate the case where signing breaks but all the tests are passing.

PR #832, opened 2020-02-12
798 payment profiles rework

rhlsthrm:
## Checklist:


.eslintrc.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     "semi": ["error", "always"],
     "spaced-comment": "off",
     "no-prototype-builtins": "off",
-    "sort-keys": ["error"],
+    "sort-keys": ["off"],
```
LayneHaber: got tired of it eh

modules/node/src/appRegistry/appRegistry.service.ts
```diff
@@ -41,25 +43,103 @@ export class AppRegistryService {
...
+      );
+      if (tx) {
+        await tx.wait();
+      }
+      try {
+        await this.cfCoreService.installApp(appInstanceId);
+      } catch (e) {
+        logger.warn(`App install failed, . Error: ${e.stack || e.message}`);
+        await this.cfCoreService.rejectInstallApp(appInstanceId);
+      }
```
LayneHaber: won't this still call `this.runPostInstallTasks` after the app is rejected?

modules/node/src/appRegistry/appRegistry.service.ts
```diff
@@ -41,25 +43,103 @@ export class AppRegistryService {
...
+          proposeInstallParams.initiatorDepositTokenAddress,
+          bigNumberify(proposeInstallParams.initiatorDeposit),
+          appInstanceId,
+          initialState.linkedHash,
+          initialState.paymentId,
+          proposeInstallParams.meta,
+        );
+        logger.debug(`Linked transfer saved!`);
+        break;
+      // TODO: add something for swap app? maybe for history preserving reasons.
```
LayneHaber: we could store swap rates used and provided?

modules/node/src/channel/channel.provider.ts
```diff
@@ -71,52 +68,43 @@ class ChannelMessaging extends AbstractMessagingProvider {
 
   async requestCollateral(pubId: string, data: { assetId?: string }): Promise<CFCoreTypes.DepositResult> {
     // do not allow clients to specify an amount to collateralize with
-    return this.channelService.requestCollateral(pubId, getAddress(data.assetId));
+    return (this.channelService.rebalance(
+      pubId,
+      getAddress(data.assetId),
+      RebalanceType.COLLATERALIZE,
+    ) as unknown) as CFCoreTypes.DepositResult;
```
LayneHaber: should this be `await`ed?

modules/node/src/channel/channel.provider.ts
```diff
@@ -71,52 +68,43 @@ class ChannelMessaging extends AbstractMessagingProvider {
...
-    return convert.PaymentProfile("str", {
-      amountToCollateralize,
-      assetId,
-      minimumMaintainedCollateral,
-    });
   }
 
-  async getPaymentProfile(pubId: string, data: { assetId?: string }): Promise<PaymentProfileRes | undefined> {
-    const prof = await this.channelRepository.getPaymentProfileForChannelAndToken(pubId, data.assetId);
+  async getRebalanceProfile(pubId: string, data: { assetId?: string }): Promise<RebalanceProfile> {
```
LayneHaber: should return `Promise<RebalanceProfile | undefined>`

modules/node/src/channel/channel.service.spec.ts
```diff
@@ -124,6 +124,7 @@ describe.skip("Channel Service", () => {
   });
 
   it("should add deposits to the onchain transaction table", async () => {
-    const result = await channelService.deposit(mkAddress("0xAAA"), One);
+    const result = await channelService.rebalance(mkXpub("0xXpub"), AddressZero, RebalanceType.COLLATERALIZE, One);
+    console.log("result: ", result);
```
LayneHaber: can probably remove

modules/node/src/onchainTransactions/onchainTransaction.service.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Injectable } from "@nestjs/common";
+import { TransactionResponse } from "ethers/providers";
+
+import { Channel } from "../channel/channel.entity";
+import { ConfigService } from "../config/config.service";
+
+import { OnchainTransactionRepository } from "./onchainTransaction.repository";
+
+@Injectable()
+export class OnchainTransactionService {
```
LayneHaber: we should have a `sendCollateral` as well right?

modules/node/src/rebalanceProfile/rebalanceProfile.repository.ts
```diff
@@ -0,0 +1,6 @@
+import { EntityRepository, Repository } from "typeorm";
+
+import { RebalanceProfile } from "./rebalanceProfile.entity";
+
+@EntityRepository(RebalanceProfile)
+export class RebalanceProfileRepository extends Repository<RebalanceProfile> {}
```
LayneHaber: do we need this?

modules/node/src/channel/channel.provider.ts
```diff
@@ -71,52 +68,43 @@ class ChannelMessaging extends AbstractMessagingProvider {
...
+    const { upperBoundReclaim, lowerBoundReclaim, upperBoundCollateralize, lowerBoundCollateralize, assetId } = prof;
+    return {
       assetId,
-      minimumMaintainedCollateral,
-    });
+      lowerBoundCollateralize: lowerBoundCollateralize.toString(),
+      lowerBoundReclaim: lowerBoundReclaim.toString(),
+      upperBoundCollateralize: upperBoundCollateralize.toString(),
+      upperBoundReclaim: upperBoundReclaim.toString(),
+    };
```
LayneHaber: could just use the conversion function

modules/node/src/config/config.service.ts
```diff
@@ -236,24 +236,32 @@ export class ConfigService implements OnModuleInit {
     return this.get(`INDRA_REDIS_URL`);
   }
 
-  async getDefaultPaymentProfile(assetId: string = AddressZero): Promise<PaymentProfile | undefined> {
+  getRebalancingServiceUrl(): string | undefined {
+    return this.get(`INDRA_REBALANCING_SERVICE_URL`);
+  }
```
LayneHaber: is this env var in the startup scripts anywhere?

should be optional

modules/node/src/appRegistry/appRegistry.service.ts
rhlsthrm: Interesting idea... I'll write it up as a separate issue.

modules/node/src/appRegistry/appRegistry.service.ts
rhlsthrm: Yikes, need a return there. Good catch.

modules/node/src/channel/channel.provider.ts
rhlsthrm: You're right, not sure why Typescript isn't catching these things.

modules/node/src/channel/channel.provider.ts
rhlsthrm: You don't technically need a to await if you're returning it since it's still a promise. There's actually a linter rule to explicitly _not_ `return await`: https://eslint.org/docs/rules/no-return-await

Now that I'm looking at that link, it actually makes performance worse to do what we are doing, oops. I'll write this as another issue, yikes.

modules/node/src/onchainTransactions/onchainTransaction.service.ts
rhlsthrm: We collateralize using the `deposit` function so we don't have/need a service level method for it.

modules/node/src/rebalanceProfile/rebalanceProfile.repository.ts
rhlsthrm: Yeah, just to keep things consistent.

modules/node/src/channel/channel.provider.ts
rhlsthrm: We've talked about deprecating the conversion functions due to complexity, so I didn't do it here, but you're right, may as well use it.

modules/node/src/config/config.service.ts
rhlsthrm: Yeah, that's why it's marked as possibly undefined. This PR does not add the integration of the service yet.

PR #833, opened 2020-02-12
Ops fix local rinkeby

LayneHaber:
## The Problem
Could not start a local indra instance pointed at rinkeby because of rpc url defaults.

## The Solution
Two changes:
- an env var called `INDRA_ETH_PROVIDER` is used to set the `eth_rpc_url` if the network is not ganache
- the `INDRA_ETH_NETWORK` is now set from an env var like other variables in the script instead of a command line argument

Note: starting the script this way will not be backwards compatible, I updated the documentation in the `README.md` where the old way was referenced, and added a section in the root `README.md`

## Checklist:


README.md
```diff
@@ -209,53 +225,55 @@ ssh -i ~/.ssh/connext-aws ubuntu@SERVER_IP bash indra/ops/collateralize.sh 1000
...
- - anything else: serves the daicard html/css/js files
+- `/api/rinkeby/hub` -> `https://rinkeby.hub.connext.network/api/hub`
+- `/api/rinkeby/eth` -> `https://rinkeby.hub.connext.network/api/eth`
+- `/api/mainnet/hub` -> `https://hub.connext.network/api/hub`
+- `/api/mainnet/eth` -> `https://hub.connext.network/api/eth`
+- anything else: serves the daicard html/css/js files
 
 ### Hub API
 
  1. AuthApiService
```
rhlsthrm: Um, not relayed to the PR but these aren't correct there aren't POST and GET methods for these...

PR #848, opened 2020-02-18
WIP Add SimpleSignatureTransferApp

balasan:
## The Solution
An app that requires the recipient to submit some signed data. If the signature matches the 'signer' parameter set by sender the payment is unlocked.

## Checklist:

Progress:
 - [x] Added Solidity App & respective tests
 - [ ] TODO: Edit Conditional and Resolve Controllers
 - [ ] TODO: Add Documentation

Questions: 
- [ ] check variable names - is data to vague? 
- [ ] do we need to be including all of the parameters in the message below? 
```
    bytes32 hash = keccak256(abi.encodePacked(
      state.amount,
      state.assetId,
      state.paymentId,
      state.data
    ));
```


rhlsthrm:
@balasan I think this can be closed since you will be exploring the FastSignedTransferApp that's currently in staging.

PR #851, opened 2020-02-19
[WIP] Dispute progression lockdown

LayneHaber:
## The Problem
`RespondToChallenge` was whack

## The Solution
Enforce that challenges can only be set with either `setState` or `setStateWithAction`. All challenges must be responded to using `respondToChallenge`.

All of the tests will fail so DO NOT MERGE YET but creating a PR to kickoff review process.

Major changes to focus on:
- `MixinSetState` and `MixinSetStateWithAction` can only be used if an outcome is set on a challenge (ie. you are restarting one), or if the challenge is fresh. major changes are in the status require -- double check that there are no weird reentrancy bugs here
- `MixinRespondToChallenge` looks completely different now, so make sure that all of this is okay. There should be more comments in there about why it was done in that way, so please check that as well.

## Checklist:


modules/contracts/contracts/adjudicator/mixins/MixinRespondToChallenge.sol
```diff
@@ -34,29 +32,141 @@ contract MixinRespondToChallenge is LibStateChannelApp, LibAppCaller, MChallenge
...
+          // Doing otherwise could violate the signers intention. For
+          // example:
+          // Signer may be fine signing a very small timeout for a favorable
+          // state. ("I made the last move in this state, so I'll win if i
+          // finalized!"). Then counterparty applies an action to it, and now 
+          // signer has very little time to react to this potentially
+          // unfavorable state. ("Now they made the last move and will win!")
+          // instead use the default timeout.
+          uint256 finalizesAt = block.number + appIdentity.defaultTimeout;
+          require(finalizesAt >= appIdentity.defaultTimeout, "uint248 addition overflow");
```
rhlsthrm: Seems like we could use safeMath to implicitly have these overflow checks, but I guess it's fine.

LayneHaber: im not sure why were not using safe math tbh, it seems like a no brainer. i was just trying to be consistent with the other contracts

PR #854, opened 2020-02-20
[node, client, types, store] Store refactor

rhlsthrm:
## Checklist:


modules/types/src/store.ts
```diff
@@ -59,13 +63,38 @@ export interface FileStorageOptions {
...
+  saveCommitment(commitmentHash: string, commitment: any[]): Promise<void>;
+  getWithdrawalCommitment(
+    multisigAddress: string,
+  ): Promise<CFCoreTypes.MinimalTransaction | undefined>;
+  saveWithdrawalCommitment(
+    multisigAddress: string,
+    commitment: CFCoreTypes.MinimalTransaction,
+  ): Promise<void>;
+  getExtendedPrvKey(): Promise<string>;
+  saveExtendedPrvKey(extendedPrvKey: string): Promise<void>;
```
bohendo: Are these being used? I'd prefer we remove these methods, it's more secure to keep private keys in memory & not ever write them to disk

modules/node/src/config/config.service.ts
```diff
@@ -120,6 +120,7 @@ export class ConfigService implements OnModuleInit {
 
     if (currentChainId !== 1) {
       const tokenConfig = await this.getTestnetTokenConfig();
+      console.log('tokenConfig: ', tokenConfig);
```
bohendo: `console.log` are great for debugging but let's either remove them or replace w proper logging system before merging

PR #855, opened 2020-02-21
WIP SimpleSignatureTransferApp add to appRegistry and to client

balasan:
This is WIP, don't merge, for reference only

## The Problem
This is the beginning of the process of adding a new Transfer app to the indra node. The current code is quite messy, untested and points to a need for a deeper refactor that could enable a more clear and canonical process for adding new apps to a Connext node. 

## Refactor ideas / things to consider
- DB schemas for different apps
  - The easiest thing to do for now was to simply add an additional app state field to the `LinkedTransfer` `transfer.entity.ts`. Does it make sense to have a more general app schema that can be used by multiple apps? It may make more sense to have a dedicated schema for each app.
  - In the case of separate schemas for each app, I would break up the schemas in to separate files to make them easier to manage. 
- appRegistry.service.ts 
  - app validators should be simple out into their own files and common code extracted into common utility modules
- appProposals.ts
  - similiar to above - we should extract the validators into app-specific files
- idea / suggestion - an `apps module` that holds all the app logic that `node` and `client` can import. alternately an apps folder inside node module?
  -  this way appRegistry, appProposals  and controllers can simply import the necessary logic for each app

## Checklist:


modules/node/src/appRegistry/appRegistry.service.ts
```diff
@@ -289,6 +319,146 @@ export class AppRegistryService implements OnModuleInit {
     logger.log(`Exchange amounts are within ${ALLOWED_DISCREPANCY_PCT}% of our rate ${ourRate}`);
   }
 
+  // TODO most of the code is duplicated from validateSimpleLinkedTransfer
+  // should refactor this into single validation method based on condition type
```
rhlsthrm: We could have a common `validateTransfer` that encapsulates all the validation general transfer apps need.

balasan:
I got started on adding the SignatureTransferApp to the client today and hit a another big pain point untangling the [types/inputs](https://github.com/ConnextProject/indra/blob/staging/modules/types/src/inputs.ts) file. The combination of hard-to distinguish type names, multiple apps & and intersection types made it particularly treacherous. 

This really should be a 20-minute copy paste + make some minor edits kind of process. I made the first refactor step and put all the SimpleTransfer input types [here](https://github.com/balasan/indra/tree/feature/signature-transfer/modules/types/src/app). 

Overall, I have a pretty strong conviction that all app-related logic, validation, types and tests should live in inside a single folder dedicated to that app. This can makes adding an app much more straight forward: 
- copy a folder of an existing template app and modify files inside
- edit .sol & tests
- edit types
- edit validators
- edit controllers
- edit integration tests

At the moment is very confusing trying making sense of files organized by functionality as opposed to app namespace across multiple modules.

Here is my experience of adding a new app so far:
- edit .sol & tests <-- all good ~ 3h for something custom
- add app to registry ~ 6+h!
  - edit `types/app`
  - edit `types/convert` - uh I need to edit more types!?!?
  - edit `client/types.ts` - not sure why this file is needed...
  - edit `node/appRegistry.service` (confusing)
  - edit `client/validation/appProposals` - more validators?
  - `node/transfer/transfer.entity` & do db migration! <-- this was a big time sync, should try to avoid this
- get started on client code ~ 4h
  - edit `types/inputs` - there goes 4h 😳

TODOs:
 - edit the resolvers (probably 2-4h with a light refactor)
 - see if any of the DB interactions need to be edited (a little scared of this one)
 - add integration tests (looks straight forward enough)

rhlsthrm:
@balasan, I like your ideas here to make folders for each app. I'm working on implementing an app myself, so I will make these re-org changes in my branch.

Issue #873, opened 2020-02-28
[client, node] Pass async transfer sender in `RECEIVE_TRANSFER_FINISHED` event data


rhlsthrm:
Currently, it is not possible to get the payment sender from the RECEIVE_TRANSFER_FINISHED event data. This is because the node is the one who sends this half of the payment.

Issue #870, opened 2020-02-27
[node] Make storing of encrypted preimage atomic


rhlsthrm:
Currently, the encrypted preimage on transfers is sent from the sender to the node as an out-of-protocol NATS message. This can cause issues if the call fails, since no one is storing the preimage at this point. We should add this to the payment meta since the node can pull it out and store it.

PR #874, opened 2020-02-28
870 store encypted preimage atomically

rhlsthrm:
## The Problem
Fixes a couple of issues, #870 and #873.

## Checklist:

### DIFF
--- a/modules/client/src/controllers/ConditionalTransferController.ts
+++ b/modules/client/src/controllers/ConditionalTransferController.ts
@@ -88,6 +88,7 @@ export class ConditionalTransferController extends AbstractController {
 
     // add encrypted preImage to meta so node can store it in the DB
     params.meta["encryptedPreImage"] = encryptedPreImage;
+    params.meta["recipient"] = recipient;
 
     // wait for linked transfer
     const ret = await this.handleLinkedTransfers({
--- a/modules/client/src/controllers/ConditionalTransferController.ts
+++ b/modules/client/src/controllers/ConditionalTransferController.ts
@@ -31,11 +31,15 @@ import {
 import { AbstractController } from "./AbstractController";
 
 type ConditionalExecutors = {
-  [index in TransferCondition]: (params: ConditionalTransferParameters) => Promise<ConditionalTransferResponse>;
+  [index in TransferCondition]: (
+    params: ConditionalTransferParameters,
+  ) => Promise<ConditionalTransferResponse>;
 };
 
 export class ConditionalTransferController extends AbstractController {
-  public conditionalTransfer = async (params: ConditionalTransferParameters): Promise<ConditionalTransferResponse> => {
+  public conditionalTransfer = async (
+    params: ConditionalTransferParameters,
+  ): Promise<ConditionalTransferResponse> => {
     this.log.info(
       `Generating conditional transfer of ${formatEther(params.amount)} ${
         params.assetId === AddressZero ? "ETH" : "Tokens"
@@ -53,10 +57,15 @@ export class ConditionalTransferController extends AbstractController {
   private handleLinkedTransferToRecipient = async (
     params: LinkedTransferToRecipientParameters,
   ): Promise<LinkedTransferToRecipientResponse> => {
-    const { amount, assetId, paymentId, preImage, recipient, meta } = convert.LinkedTransferToRecipient(
-      `bignumber`,
-      params,
-    );
+    params.meta = params.meta && typeof params.meta === "object" ? params.meta : {};
+    const {
+      amount,
+      assetId,
+      paymentId,
+      preImage,
+      recipient,
+      meta,
+    } = convert.LinkedTransferToRecipient(`bignumber`, params);
 
     const freeBalance = await this.connext.getFreeBalance(assetId);
     const preTransferBal = freeBalance[this.connext.freeBalanceAddress];
@@ -69,21 +78,22 @@ export class ConditionalTransferController extends AbstractController {
       invalidXpub(recipient),
     );
 
-    const linkedHash = createLinkedHash(amount, assetId, paymentId, preImage);
+    // set recipient and encrypted pre-image on linked transfer
+    // TODO: use app path instead?
+    const recipientPublicKey = fromExtendedKey(recipient).derivePath(`0`).publicKey;
+    const encryptedPreImage = await encryptWithPublicKey(
+      recipientPublicKey.replace(/^0x/, ``),
+      preImage,
+    );
+
+    // add encrypted preImage to meta so node can store it in the DB
+    params.meta["encryptedPreImage"] = encryptedPreImage;
 
     // wait for linked transfer
     const ret = await this.handleLinkedTransfers({
       ...params,
       conditionType: LINKED_TRANSFER,
     });
-
-    // set recipient and encrypted pre-image on linked transfer
-    // TODO: use app path instead?
-    const recipientPublicKey = fromExtendedKey(recipient).derivePath(`0`).publicKey;
-    const encryptedPreImage = await encryptWithPublicKey(recipientPublicKey.replace(/^0x/, ``), preImage);
-    // TODO: if this fails for ANY REASON, uninstall the app to make sure that
-    // the sender doesnt lose any money
-    await this.connext.setRecipientAndEncryptedPreImageForLinkedTransfer(recipient, encryptedPreImage, linkedHash);
 
     // publish encrypted secret
     // TODO: should we move this to its own file?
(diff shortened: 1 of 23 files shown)


modules/client/src/controllers/ConditionalTransferController.ts
```diff
@@ -69,22 +78,24 @@ export class ConditionalTransferController extends AbstractController {
       invalidXpub(recipient),
     );
 
-    const linkedHash = createLinkedHash(amount, assetId, paymentId, preImage);
+    // set recipient and encrypted pre-image on linked transfer
+    // TODO: use app path instead?
```
LayneHaber: Theres no great way to allow the recipient and sender to understand what the app key should be, since they don't know what nonce the other would be at or be expecting. I think leaving it as the free balance addr is best

modules/client/src/controllers/ConditionalTransferController.ts
```diff
@@ -165,7 +187,12 @@ export class ConditionalTransferController extends AbstractController {
     appInfo: DefaultApp,
     meta?: object,
   ): Promise<string | undefined> => {
-    const { appDefinitionAddress: appDefinition, outcomeType, stateEncoding, actionEncoding } = appInfo;
+    const {
+      appDefinitionAddress: appDefinition,
+      outcomeType,
+      stateEncoding,
+      actionEncoding,
+    } = appInfo;
```
LayneHaber: i really thought i got these things with my linting PR 😢

modules/client/src/controllers/ResolveConditionController.ts
```diff
@@ -207,8 +212,9 @@ export class ResolveConditionController extends AbstractController {
 
     return {
       appId,
-      freeBalance: await this.connext.getFreeBalance(assetId),
+      sender,
       paymentId,
+      meta,
```
LayneHaber: i feel like we should keep in the free balance to make this change purely additive instead of breaking

modules/node/src/listener/listener.service.ts
```diff
@@ -134,26 +134,53 @@ export default class ListenerService implements OnModuleInit {
...
+            `Got update state event for a receiver's transfer app (transfer.id: ${transfer.id}) with unexpected status: ${transfer.status}`,
+          );
+          return;
+        }
+
+        // transfers are set to `PENDING` when created. They are set to
+        // `FAILED` when the receiver rejects an install event. If a transfer
+        // makes it to the `UPDATE_STATE_EVENT` portion, it means the transfer
+        // was successfully installed. There is no reason to not redeem it in
+        // that case.
```
LayneHaber: we could probably also just make the condition
```typescript
if (transfer.status !== LinkedTransferStatus.PENDING) { ... }
```

but we don't technically *have* to

modules/node/src/transfer/transfer.provider.ts
```diff
@@ -45,7 +45,7 @@ export class TransferMessaging extends AbstractMessagingProvider {
     return await this.transferService.resolveLinkedTransfer(pubId, paymentId, linkedHash);
   }
 
-  // TODO: types
+  // DEPRECATED
```
LayneHaber: can we remove?

LayneHaber: I don't think this was ever used externally

modules/node/src/transfer/transfer.repository.ts
```diff
@@ -16,7 +16,7 @@ export class PeerToPeerTransferRepository extends Repository<PeerToPeerTransfer>
 @EntityRepository(LinkedTransfer)
 export class LinkedTransferRepository extends Repository<LinkedTransfer> {
   async findByPaymentId(paymentId: string): Promise<LinkedTransfer | undefined> {
-    return await this.findOne({ where: { paymentId } });
+    return await this.findOne({ relations: ["senderChannel"], where: { paymentId } });
```
LayneHaber: what does this query change do?

modules/node/src/transfer/transfer.service.ts
```diff
@@ -121,14 +128,15 @@ export class TransferService {
...
   async setRecipientAndEncryptedPreImageOnLinkedTransfer(
     senderPublicIdentifier: string,
     recipientPublicIdentifier: string,
     encryptedPreImage: string,
     linkedHash: string,
   ): Promise<LinkedTransfer> {
-    this.log.debug(
-      `setRecipientAndEncryptedPreImageOnLinkedTransfer(${senderPublicIdentifier}, ${recipientPublicIdentifier}, ${encryptedPreImage}, ${linkedHash}`,
+    this.log.warn(
+      `setRecipientAndEncryptedPreImageOnLinkedTransfer is DEPRECATED, client needs to be updated`,
```
LayneHaber: this should be logged client side somewhere instead

sidetone -- will this change break all our existing clients who we want to reclaim from? (i don't think so, want to make sure)

modules/test-runner/src/transfer/asyncTransferOffline.test.ts
```diff
@@ -94,7 +106,8 @@ describe("Async transfer offline tests", () => {
    * Avoid `transfer.set-recipient.${this.userPublicIdentifier}` endpoint
    * with the hub. The hub should clean these up during disputes
    */
-  it("sender successfully installs transfer, goes offline before sending paymentId or preimage, then comes online and has the pending installed transfer", async () => {
+  // TODO: this doesn't work bc `transfer.set-recipient` isn't used anymore
+  it.skip("sender successfully installs transfer, goes offline before sending paymentId or preimage, then comes online and has the pending installed transfer", async () => {
```
LayneHaber: i dont think that we need this test anymore -- as soon as the sender has the app installed, it should be okay. theres no way for this failure to occur, and the positive case has been tested when the sender is offline and the receiver comes back online

modules/test-runner/src/transfer/asyncTransferOffline.test.ts
```diff
@@ -132,7 +145,7 @@ describe("Async transfer offline tests", () => {
    *
    * Recipient should be able to claim payment regardless.
    */
-  it("sender successfully installs transfer, goes offline before sending paymentId/preimage, and stays offline", async () => {
+  it.skip("sender successfully installs transfer, goes offline before sending paymentId/preimage, and stays offline", async () => {
```
LayneHaber: is there a reason this is skipped or was it just causing problems?

modules/client/src/controllers/ResolveConditionController.ts
rhlsthrm: I really don't like the pattern of returning free balance like this so I'm trying to remove it. I think now is the best time to break things like this.

modules/node/src/transfer/transfer.repository.ts
rhlsthrm: Adds the `senderChannel` as an object so you can query `transfer.senderChannel.publicIdentifier` for example.

modules/node/src/transfer/transfer.provider.ts
rhlsthrm: Yes I would love to.

modules/node/src/transfer/transfer.service.ts
rhlsthrm: Nope, if they have ever called this in the past, everything will be set for them in the DB.

modules/test-runner/src/transfer/asyncTransferOffline.test.ts
rhlsthrm: Perfect I love deleting stuff lol.

modules/test-runner/src/transfer/asyncTransferOffline.test.ts
rhlsthrm: This is one that was causing issues that I wanted help with. We should probably unskip it and fix it.

modules/node/src/transfer/transfer.provider.ts
rhlsthrm: This is a breaking change (old clients won't be able to transfer with this, but we can bump major version). That's why we should put other breaking changes into this PR.

modules/test-runner/src/transfer/asyncTransferOffline.test.ts
rhlsthrm: Done, now we have to deal with it.

modules/node/src/transfer/transfer.service.ts
```diff
@@ -369,14 +369,14 @@ export class TransferService {
...
 
-  // TODO: why don't these functions throw errors, this breaks the patterns we normally use
   async reclaimLinkedTransferCollateralByAppInstanceIdIfExists(
     appInstanceId: string,
   ): Promise<void> {
     const transfer = await this.linkedTransferRepository.findByReceiverAppInstanceId(appInstanceId);
     if (!transfer || transfer.status !== LinkedTransferStatus.REDEEMED) {
-      this.log.debug(`Did not find transfer`);
-      return;
+      throw new Error(
```
rhlsthrm: But then this will throw errors every time an app is uninstalled right? Since we call this in the listener.

LayneHaber: handled these errors there

PR #885, opened 2020-03-11
support nats for defaults

pedrouid:
## Checklist:


modules/client/src/lib/default.ts
```diff
@@ -57,9 +61,11 @@ export async function getDefaultOptions(
     throw new Error(`Provided network (${network.toLowerCase()}) is not supported`);
   }
 
+  const nodeUrlProtocol = isNode ? "nats" : "wss";
+
   const urlOptions = {
     ethProviderUrl: `https://${baseUrl}/ethprovider`,
```
rhlsthrm: I feel like by default we should use the ethers default providers that provide fallbacks and other useful stuff.

PR #907, opened 2020-03-19
improve crypto module signing and recovery with tests

pedrouid:
## Checklist:


modules/client/src/channelProvider.ts
```diff
@@ -87,8 +87,8 @@ export class CFCoreRpcConnection extends ConnextEventEmitter implements IRpcConn
         result = await this.signWithdrawCommitment(params.message);
         break;
       case chan_nodeAuth:
```
rhlsthrm: This is renamed to `signDigest` and consolidated with `signWithdrawalCommitment` in the NATS PR. Would be good to see how it's done there.

pedrouid: well basically signWithdrawlCommitment was already signDigest anyway so it wont be a problem to converge the two

PR #910, opened 2020-03-22
Replace signDigestWithEthers with signDigest from crypto module + other improvements

pedrouid:
requires #914


modules/cf-core/src/store_.ts
```diff
@@ -0,0 +1,272 @@
+import { CriticalStateChannelAddresses, AppInstanceJson, StateChannelJSON } from "@connext/types";
```
LayneHaber: we can delete this

PR #914, opened 2020-03-23
(1) Signing Refactor - Hashing and Signature parsing improvements 

pedrouid:
This PR is mostly a preparation PR for a following signing refactor. Basically we currently sign digests and handle digests as the data to sign. This is bad practice for the following reasons:
1. we hash unnecessarily to get the data to sign
2. we expose the wallet signDigest functionality

The first problem is using the commitment hashes as the data for signing which means that we care calling keccak256 simply to evaluate the commitment data even without signing.
Solution is to simply encode the commitment parameters with solidityPack and use them as the data to sign and evaluate.

The second problem is the most critical, since we have auto-signing capabilities in our clients, we are currently exposing the signDigest functionality which will basically allow the wallet to auto-sign any secp256k1 digests.
Solution is to create a new method that uses a Channel-specific hashing prior to signing, therefore the data to be signed can only be used for our use-cases.

Additionally this PR removes some unnecessary mapping that parses Signatures and instead expect them to be formatted as encoded byte strigns


Makefile
```diff
@@ -328,7 +328,7 @@ messaging: node-modules types $(shell find $(messaging)/src $(find_options))
 	$(docker_run) "cd modules/messaging && npm run build"
 	$(log_finish) && mv -f $(totalTime) $(flags)/$@
 
-node: cf-core contracts types apps messaging $(shell find $(node)/src $(node)/migrations $(find_options))
+node: cf-core types contracts crypto apps messaging $(shell find $(node)/src $(node)/migrations $(find_options))
```
LayneHaber: if the `node` rule needs the `crypto` dep, then shouldn't the client? and since its a `cf-core` dep, doesn't that mean that including `cf-core` will force the `crypto` module to get built anyways?

modules/cf-core/src/ethereum/conditional-transaction-commitment.ts
```diff
@@ -23,7 +23,7 @@ export class ConditionalTransactionCommitment extends MultisigCommitment {
     public readonly freeBalanceAppIdentityHash: string,
     public readonly interpreterAddr: string,
     public readonly interpreterParams: string,
-    participantSignatures: Signature[] = [],
+    participantSignatures: string[] = [],
```
LayneHaber: nice

Makefile
pedrouid: definitely we should also cleanup duplicated `types` build commands

PR #917, opened 2020-03-23
(3) Signing Refactor - Refactor crypto tests and make recoverAddress async

pedrouid:
requires #910


modules/cf-core/src/machine/middleware.ts
```diff
@@ -18,8 +18,8 @@ export class MiddlewareContainer {
   public async run(opCode: Opcode, args: any[]) {
     const middleware = this.middlewares[opCode][0];
 
-    if (middleware === undefined) {
-      throw Error(`Attempted to run middleware for opcode ${opCode} but none existed`);
+    if (typeof middleware === "undefined") {
+      throw new Error(`Attempted to run middleware for opcode ${opCode} but none existed`);
```
LayneHaber: this won't catch nulls

modules/cf-core/src/machine/protocol-runner.ts
```diff
@@ -76,8 +76,8 @@ export class ProtocolRunner {
   public async runProtocolWithMessage(msg: ProtocolMessage) {
     const protocol = getProtocolFromName(msg.protocol);
     const step = protocol[msg.seq];
-    if (step === undefined) {
-      throw Error(`Received invalid seq ${msg.seq} for protocol ${msg.protocol}`);
+    if (typeof step === "undefined") {
+      throw new Error(`Received invalid seq ${msg.seq} for protocol ${msg.protocol}`);
```
LayneHaber: same as above

modules/cf-core/src/methods/app-instance/take-action/controller.ts
```diff
@@ -38,7 +38,7 @@ export default class TakeActionController extends NodeController {
     const { appInstanceId, action } = params;
 
     if (!appInstanceId) {
-      throw Error(NO_APP_INSTANCE_FOR_TAKE_ACTION);
+      throw new Error(NO_APP_INSTANCE_FOR_TAKE_ACTION);
```
LayneHaber: this looks so tedious im sorry lol

modules/cf-core/src/machine/middleware.ts
pedrouid: but the previous one wouldnt either
```javascript
const testNull = null

null === undefined
> false

typeof testNull === 'undefined'
> false
```

modules/cf-core/src/methods/app-instance/take-action/controller.ts
pedrouid: nah it was just global replace

modules/cf-core/src/machine/protocol-runner.ts
pedrouid: read comment above

PR #921, opened 2020-03-25
🐛  Get hashlock transfers improvements and fixes

rhlsthrm:
## The Problem

There were issues with resolving hashlock transfers as well as determining statuses. All these have been fixed.

## Checklist:

### DIFF
--- a/modules/node/src/appRegistry/appActions.service.ts
+++ b/modules/node/src/appRegistry/appActions.service.ts
@@ -5,7 +5,6 @@ import {
   AppAction,
   convertHashLockTransferAppState,
   convertFastSignedTransferAppState,
-  convertLinkedTransferAppState,
 } from "@connext/apps";
 import {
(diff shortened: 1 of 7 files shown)


modules/node/src/appRegistry/appActions.service.ts
```diff
@@ -226,6 +225,7 @@ export class AppActionsService {
     } as HashLockTransferAppAction);
 
     await this.cfCoreService.uninstallApp(senderApp.identityHash);
-    this.log.info(`Reclaimed collateral from ${senderApp.identityHash}`);
+    console.log(`UNINSTALLED SENDER APP AT ${Date.now()}`);
```
pedrouid: typo?

rhlsthrm: Ah this is for the test, should be deleted.

PR #941, opened 2020-03-30
934 init haproxy

bohendo:
## The Problem

nginx is a subpar proxy, it has trouble keeping the ws connection alive

## The Solution

use a better one: haproxy

## Checklist:


kthomas:
This PR is a lot bigger than just haproxy 🤣 

That aside, the haproxy bits look good so far. How have you tested this?

ops/proxy/indra/prod.cfg
```diff
@@ -0,0 +1,67 @@
...
+  option httplog
+  option forwardfor
+  option redispatch
+  timeout connect 3000    # 3 seconds
+  timeout client  300000  # 5 minutes
+  timeout server  300000  # 5 minutes
+
+frontend http_443
+  bind *:443 ssl crt fullchain.pem
+  mode http
```
kthomas: This should be `tcp` for the NATS websocket frontend.

bohendo:
> How have you tested this?
Normal tests that make sure everything works both in browser (ws) and between two node processes (tcp). No load tests or anything more specific yet (eg not sure how to test keep-alive stuff)

bohendo:
> How have you tested this?
Normal tests that make sure everything works both in browser (ws) and between two node processes (tcp). No load tests or anything more specific yet (eg not sure how to test keep-alive stuff)

Also, what's the difference between connection to the nats ws port 4221 via an http-mode proxy vs a tcp-mode one (isn't ws on layer 7 just like http?). You'll see I'm currently doing both in my indra proxy config

PR #952, opened 2020-04-01
950 postgres client store

rhlsthrm:
## Checklist:


modules/store/src/connextStore.ts
```diff
@@ -60,8 +63,24 @@ export class ConnextStore implements IClientStore {
...
-      case StoreTypes.File:
+      case StoreTypes.Postgres: {
+        this.internalStore = new KeyValueStorage(
+          (opts.storage as WrappedPostgresStorage) ||
+            new WrappedPostgresStorage(
+              this.prefix,
+              this.separator,
+              DEFAULT_POSTGRES_TABLE_NAME,
+              opts.sequelize,
+              opts.postgresConnectionUri,
```
LayneHaber, line 76: client stores usually have defaults -- are we ignoring those with the postgres one? If so, it should probably error out at some point before trying to create the store

modules/store/src/wrappers/keyValueStorage.ts
```diff
@@ -96,11 +92,11 @@ export class KeyValueStorage implements WrappedStorage, IClientStore {
   async getStateChannel(multisigAddress: string): Promise<StateChannelJSON | undefined> {
     const channelKey = this.getKey(CHANNEL_KEY, multisigAddress);
     const item = await this.getItem(channelKey);
-    const chan = safeJsonParse(item);
-    if (!chan) {
+    console.log('item: ', item);
```
LayneHaber, line 95: can remove once tests pass

modules/test-runner/src/store/store.test.ts
```diff
@@ -12,11 +12,15 @@ import {
   TEST_STORE_CONDITIONAL_COMMITMENT,
 } from "../util";
 
+export const storeTypesNoPostgres = Object.keys(StoreTypes).filter(
+  store => store !== StoreTypes.Postgres,
+);
+
```
LayneHaber, line 18: is there a way to test this with a postgres store?

modules/store/src/connextStore.ts
rhlsthrm: It does error in the wrapped version.

modules/test-runner/src/store/store.test.ts
rhlsthrm: Yes, I figured it out, will add this to the suite.

modules/store/src/connextStore.ts
```diff
@@ -60,8 +63,24 @@ export class ConnextStore implements IClientStore {
           ),
         );
         break;
+      }
 
-      case StoreTypes.File:
+      case StoreTypes.Database: {
```
pedrouid, line 68: I feel like Sequelize is a better name for this. Because we can have other database storage wrappers

modules/store/src/connextStore.ts
```diff
@@ -60,8 +63,24 @@ export class ConnextStore implements IClientStore {
...
+        this.internalStore = new KeyValueStorage(
+          (opts.storage as WrappedDatabaseStorage) ||
+            new WrappedDatabaseStorage(
```
pedrouid, line 71: Same here, I would call `SequelizeSotrage`

modules/store/src/wrappers/fileStorage.ts
```diff
@@ -49,12 +51,13 @@ export class FileStorage implements WrappedStorage {
     return path.join(this.fileDir, fileName);
   }
 
-  async getItem(key: string): Promise<string | undefined> {
+  async getItem(key: string): Promise<any | undefined> {
```
pedrouid, line 54: Perhaps we could also improve these `any` types with type injection

rhlsthrm: That's what I was attempting but it was becoming difficult. Will take another look.

modules/store/src/connextStore.ts
rhlsthrm: But it doesn't actually require Sequelize, it can be instantiated with just a Postgres connection URL.

pedrouid: yeah but that connection url is then instantiated with sequelize so technically it can be any Sequelize compatible database but let's say I want to use mongodb or faunadb, they are not Sequelize compatible

PR #958, opened 2020-04-03
955 Connext-specific signing prefix

### LINKED ISSUE #955: [crypto, contracts] Connext-specific signing prefix
To protect against using connext-related signing utils for non-Connext usecases, we need to add a Connext-specific prefix to all of our signing/recovery functions.

pedrouid:
fix #955


modules/cf-core/src/testing/scenarios/install.spec.ts
```diff
@@ -66,7 +66,11 @@ describe("Node method follows spec - install", () => {
             multisigAddress,
             CONVENTION_FOR_ETH_TOKEN_ADDRESS,
           );
-          assertProposeMessage(nodeA.publicIdentifier, msg, proposeInstallParams);
+          // FIXME: multisigAddress is undefined
+          assertProposeMessage(nodeA.publicIdentifier, msg, {
+            ...proposeInstallParams,
+            multisigAddress,
+          });
```
pedrouid, line 73: This shouldn't be merge before this is fixed first

modules/cf-core/src/testing/scenarios/install.spec.ts
```diff
@@ -150,7 +153,8 @@ describe("Node method follows spec - install", () => {
             multisigAddress,
             erc20TokenAddress,
           );
-          assertProposeMessage(nodeA.publicIdentifier, msg, proposedParams);
+          // FIXME: multisigAddress is undefined
+          assertProposeMessage(nodeA.publicIdentifier, msg, { ...proposedParams, multisigAddress });
```
pedrouid, line 157: Same scenario here...

PR #971, opened 2020-04-06
957 types fixes

ArjunBhuptani:
Implemented 
- Swap method should return a SwapResponse
- Transfer method should not return any
- CREATE_TRANSFER event name is inconsistent
- Remove CfCoreController class in node
- Remove verifyAppSequenceNumber

Not implemented
- Remove signerAddress, consolidate on freeBalanceAddress
- Remove getAppState (are we sure this is right?)
- Rename getAppInstance to getAppInstanceDetails
- Change all references to appId, appIdentityHash, appHash, etc to appInstanceId.


modules/cf-core/src/methods/app-instance/install.ts
```diff
@@ -60,20 +60,20 @@ export async function install(
   params: MethodParams.Install,
   initiatorXpub: string,
 ): Promise<AppInstanceProposal> {
-  const { appInstanceId } = params;
+  const { appIdentityHash } = params;
 
-  if (!appInstanceId || !appInstanceId.trim()) {
+  if (!appIdentityHash || !appIdentityHash.trim()) {
     throw new Error(NO_APP_INSTANCE_ID_TO_INSTALL);
```
pedrouid, line 66: Error should be renamed to `NO_APP_IDENTITY_HASH_TO_INSTALL`

modules/cf-core/src/methods/app-instance/install.ts
```diff
@@ -60,20 +60,20 @@ export async function install(
...
   }
 
-  const proposal = await store.getAppProposal(appInstanceId);
+  const proposal = await store.getAppProposal(appIdentityHash);
   if (!proposal) {
-    throw new Error(NO_PROPOSED_APP_INSTANCE_FOR_APP_INSTANCE_ID(appInstanceId));
+    throw new Error(NO_PROPOSED_APP_INSTANCE_FOR_APP_INSTANCE_ID(appIdentityHash));
```
pedrouid, line 71: same here

modules/cf-core/src/methods/app-instance/install.ts
```diff
@@ -60,20 +60,20 @@ export async function install(
...
   }
 
-  const json = await store.getStateChannelByAppInstanceId(appInstanceId);
+  const json = await store.getStateChannelByAppIdentityHash(appIdentityHash);
   if (!json) {
-    throw new Error(NO_STATE_CHANNEL_FOR_APP_INSTANCE_ID(appInstanceId));
+    throw new Error(NO_STATE_CHANNEL_FOR_APP_INSTANCE_ID(appIdentityHash));
```
pedrouid, line 76: and here

Issue #1043, opened 2020-04-20
[utils] dedup utils


bohendo:
While creating the utils module, we moved all generic utilities from other modules into one place.. but we haven't actually organized or de-duplicated these utils yet. We ought to review utils & remove any that are unused or duplicates.

PR #1047, opened 2020-04-20
1043 dedup utils

bohendo:
## The Problem

Lots of duplicate utils leftover after consolidating them in their own module.

closes issue #1043

## The Solution

dedup & clean stuff up


modules/client/src/default.ts
```diff
@@ -1,16 +1,21 @@
...
 
+const removeUndefinedFields = <T>(obj: T): T => {
+  Object.keys(obj).forEach(key => typeof obj[key] === "undefined" && delete obj[key]);
+  return obj;
+};
+
 const getGeneratedSigner = (): string => {
   if (typeof window === "undefined" || typeof window.localStorage === "undefined") {
-    return getRandomPrivateKey();
+    return hexlify(randomBytes(32));
```
pedrouid, line 14: shouldn't this use getRandomBytes32 instead

bohendo: Yea, this particular `hexlify(randomBytes` call was temporary to fix broken imports while reorganizing. I replaced it with a proper getRandom function in a later commit :+1:

bohendo: Oh wait, this was in a different spot than I thought. No I didn't fix it but I will. :facepalm:

PR #1059, opened 2020-04-23
1056 slowdown investigation

### LINKED ISSUE #1056: [ops, node] slowdown investigation
In some situations the node slows to a crawl taking several minutes to return a response to a simple get channel query. We need:

 - better debug logs on node so that we have better visibility into where it's getting stuck
 - logs on db so that we can get visibility into whether some queries are taking a long time to run and to alert us if some parts of the node are DOSing the db.

bohendo:
closes #1056 

## The Problem

ETC friends were having trouble getting Indra deployed

## The Solution

Run through a handful of little tasks to make their life easier:
 - use hardcoded swap rate in dev-mode to eliminate possibility of memory leaks due to repeated unhandled promise rejections.
 - upgrade env var management to make custom envs easier to manage
 - upgraded logging for better visibility into database activity


modules/node/src/swapRate/swapRate.service.ts
```diff
@@ -69,9 +69,12 @@ export class SwapRateService implements OnModuleInit {
       newRate = await Promise.race([
         new Promise(async (resolve, reject): Promise<void> => {
           switch (priceOracleType) {
-            case "UNISWAP":
+            case PriceOracleTypes.UNISWAP:
               resolve(await this.getUniswapRate(from, to));
               break;
+            case PriceOracleTypes.HARDCODED:
+              resolve(await this.config.getDefaultSwapRate(from, to));
```
rhlsthrm, line 76: Good for now, but I was thinking we could have a way for a custom hardcoded rate to be provided via the config, within the `allowedSwap` env var.

modules/utils/src/logger.ts
```diff
@@ -3,12 +3,12 @@ import { ILogger, ILoggerService } from "@connext/types";
 export const logTime = (log: ILogger, start: number, msg: string) => {
   const diff = Date.now() - start;
   const message = `${msg} in ${diff} ms`;
-  if (diff < 10) {
-    log.info(message);
+  if (diff < 25) {
+    log.debug(message);
```
rhlsthrm, line 7: Can we keep this as info? There's a few logs in cf-core that are being made with this which we lose with debug level.

rhlsthrm: Specifically I'm talking about in protocols, we log things like `... Started` but the `... Complete` log is in this logTime function so we would lose it which makes debugging very difficult.

rhlsthrm:
Approved pending CI.

modules/node/src/swapRate/swapRate.service.ts
bohendo: The `AllowedSwap` type can't support this w/out some changes. I agree that being able to specify a swap rate as part of env vars would be good but don't think it's worth the effort at this point.

PR #1065, opened 2020-04-24
Change BN.js version

pedrouid:
Most (if not all) packages depend on bn.js v4.x so installing v5.x is redudant since some packages like `chain-bn` is dependening on the bn.js installed by ethers.js


pedrouid:
replaced by #1067

PR #1072, opened 2020-04-28
[DO NOT MERGE] Experimental/multihop

rhlsthrm:
## The Problem
We need multihop!

## The Solution
Built an initial multihop implementation very naively:
- `make start-multihop-headless` spins up 2 nodes
- Added node HTTP endpoint to trigger a channel initiation with a specified recipient
- Client provides multihop route in `meta` parameter of transfer, with a key `path`
- `ConditionalTransferTypes.HashlockTransfer` transfers will check the path and transfer along that path, to simulate onion-based routing

Breaking change:
- Need to change all messaging paths to prefix with `nodeIdentifier` so that multiple clients can be talking to multiple nodes on the same messaging server.

## Checklist:


ArjunBhuptani:
Closing this as it's no longer compatible with our protocol. We can rewrite these changes in the future.

PR #1103, opened 2020-05-07
🎨 Clean up store interface, allow passing in prefix

rhlsthrm:
## Checklist:

### DIFF
--- a/modules/store/src/connextStore.ts
+++ b/modules/store/src/connextStore.ts
@@ -24,7 +24,6 @@ import {
   WrappedAsyncStorage,
   WrappedLocalStorage,
   WrappedSequelizeStorage,
-  WrappedMemoryStorage,
 } from "./wrappers";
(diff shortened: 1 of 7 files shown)


modules/store/src/connextStore.ts
```diff
@@ -83,11 +81,10 @@ export class ConnextStore implements IClientStore {
       case StoreTypes.File: {
         this.internalStore = new KeyValueStorage(
           new WrappedSequelizeStorage(
+            `sqlite:${opts.fileDir}/${storeDefaults.SQLITE_STORE_NAME}`,
```
pedrouid, line 84: Shouldn't we just remove StoreTypes.File all together? I'm pretty sure only rest-api-client was using it

rhlsthrm: `File` is basically an alias for the SQLite store. We can change the name to be more specific.

PR #1105, opened 2020-05-07
1104 transactions

### LINKED ISSUE #1104: [store] Transactions implemented
**Describe the bug**
Need to put updating items within a transaction so you can capture concurrent writes properly. Currently blocking #1063

LayneHaber:
Fixes #1104


LayneHaber:
Adds a store dep, so will need to be npm-published for cd to pass

Issue #1116, opened 2020-05-10
[contracts, client, node, apps] Move initial state validation to init() fn in appDefinition


ArjunBhuptani:
Right now, we have to create new initial state validation libraries for every new application that we write. This makes it harder to add new apps and makes the node code less easy to understand. We also have a weird pattern where a proposal can be countersigned by a responder, even if they disagree with the proposed values.

We can move these validators into an `init()` function within the appDefinition which is call from within the propose protocol. This way, any app that makes it through the propose protocol, must *necessarily* have already been validated.

LayneHaber:
Related issue for moving validation from listeners into validation: #1158

ArjunBhuptani:
Deprioritizing this until we have a need to support many more apps

PR #1117, opened 2020-05-10
1116 init validation

ArjunBhuptani:
First half of #1116 -- remaining tasks:
1. Move base/common validators and some special case ones to injected validator middleware for propose
2. Finish setting up swapApp init validators
3. Remove app/validation logic from client and node
4. Delete apps package


ArjunBhuptani:
Closing this because we're abandoning this work for now. The overhead of validating everything onchain is too high given that we're not adding lots and lots of apps yet.

PR #1126, opened 2020-05-15
Make bots work!

rhlsthrm:
## The Solution

`bash ops/test/bot.sh 3` will spin up 3 pairs of ping pong bots

## Checklist:


modules/test-runner/src/transfer/hashLockTransfer.test.ts
```diff
@@ -245,7 +245,7 @@ describe("HashLock Transfers", () => {
     } as NodeResponses.GetHashLockTransfer);
   });
 
-  it("gets a completed hashlock transfer by lock hash", async () => {
+  it.only("gets a completed hashlock transfer by lock hash", async () => {
```
bohendo, line 248: rm only

PR #1169, opened 2020-05-26
Hotfix get hashlock transfer

ArjunBhuptani:


modules/test-runner/src/transfer/hashLockTransfer.test.ts
```diff
@@ -7,8 +7,9 @@ import {
   NodeResponses,
   PublicParams,
   EventPayloads,
+  HashLockTransferAppState,
 } from "@connext/types";
-import { getRandomBytes32 } from "@connext/utils";
+import { getRandomBytes32, ConsoleLogger, stringify } from "@connext/utils";
```
bohendo, line 12: Looks like these imports were only used temporarily while debugging, let's remove them.
Test runner output goes to terminal which means next time you can try using the :rainbow: `ColorfulLogger` :rainbow:

ArjunBhuptani: oops that was actually an accidental import while trying to `console.log`

PR #1174, opened 2020-05-27
Export migrate script

rhlsthrm:
Work started to export the migrate script so that installers of this package can migrate contracts directly without installing or copy-pasting anything. Unfortunately, it isn't fully working.

# Update by Bo

We've revamped the contracts module. It now contains:
 - `src.sol` (formerly `contracts`) contains solidity source code, compiled output goes into `artifacts`
 - `src.ts` contains all typescript source code including commitments, tests, and the new cli system.

There's a new bot-style cli, source is in: `contracts/src.ts/cli.ts` and `contracts/src.ts/commands/*`
The old migration script has been dissolved into 3 commands: `migrate`, `fund`, and `snapshot`

We can now migrate contracts by running a command like:
```
npx @connext/contracts migrate --mnemonic "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat" --eth-provider "http://localhost:8545"
```


.eslintrc.js
```diff
@@ -25,6 +26,7 @@ module.exports = {
   env: {
     browser: true,
     es6: true,
+    mocha: true,
```
rhlsthrm, line 29: Wow, I had no idea you could do this.

modules/bot-registry/package.json
```diff
@@ -11,12 +11,12 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@connext/utils": "6.5.7",
-    "body-parser": "^1.19.0",
-    "express": "^4.17.1"
+    "@connext/utils": "6.6.3",
```
rhlsthrm, line 14: We should make npm-publish update this.

modules/contracts/src.ts/constants.ts
```diff
@@ -0,0 +1,59 @@
+import { Options } from "yargs";
+
+export const botMnemonics = [
```
rhlsthrm, line 3: I'm not sure we're even using these mnemonics anymore, we should just remove them.

modules/bot-registry/package.json
bohendo: Ordinarily it would update `@connext/utils` to the newly published version in all modules, but that change got lost during a merge so I had to do it manually

PR #1177, opened 2020-05-29
1176 better events

### LINKED ISSUE #1176: Use better event emitter
This one would solve a lot of problems we have been dealing with: https://github.com/garronej/evt

LayneHaber:
## The Problem
Fixes #1176 in the client, node

Additionally makes some types changes to make sure all of our types are consistent.

## Checklist:


LayneHaber:
Current status:
- cf tests are failing when the `@connext/utils` module exports the new typed emitter class
- node still needs to upgrade to use `waitFor` and `attachOnce`
- ideally would be able to remove listeners by event name, this could be done by manipulating contexts, but is not implemented

LayneHaber:
Update:
- node is fixed
- cf core tests still having issues (integration tests pass, the cf core tests pass if you remove the `typedEmitter` export from the utils package
- some functionality around removing listeners is dropped (not important rn)

PR #1193, opened 2020-06-01
[PROVISIONAL] Performance improvements - part 1

mslipper:
This PR includes the first set of performance improvements to the Indra codebase. Specifically, it contains:

- Write-through database caching for the most-called database queries.
- Memoization of often-calculated hash values in a size-limited cache structure.

Performance changes were benchmarked using a bot setup that initiates 10 payments from bot 2 to bot 1. Performance numbers for the conditional payments are as follows:

|        | Average | 95th Percentile | DB Queries |
|--------|---------|-----------------|------------|
| Before | 472.2   | 589.6           | 1115       |
| After  | 350     | 659.8           | 756        |

The increased 95th percentile latency in the "After" scenario stems from one slow payment. I believe that particular payment resulted from a collateralization operation happening while the payment was posting.

I'll make comments inline on what I've changed in this PR. I'd like to get a thumbs up/thumbs down on the general direction, and (of course) any feedback you all might have!


Makefile
```diff
@@ -105,6 +105,7 @@ quick-reset:
 	bash ops/db.sh 'truncate table onchain_transaction cascade;'
 	bash ops/db.sh 'truncate table rebalance_profile cascade;'
 	bash ops/db.sh 'truncate table app_instance cascade;'
+	bash ops/redis.sh 'flushall'
```
mslipper, line 108: Redis is used a lot more heavily now, as such it may contain stale values that need to be removed during resets.

modules/bot/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "dist/src/index.js",
   "scripts": {
     "build": "rm -rf dist && tsc --project tsconfig.json",
-    "start": "node dist/src/index.js"
+    "start": "node --inspect=0.0.0.0:9229 dist/src/index.js"
```
mslipper, line 8: It's useful to be able to profile bots directly, so this directive allows us hook them up to a profiler.

modules/bot/src/agents/bot.ts
```diff
@@ -53,8 +53,8 @@ export default {
     const nodeUrl = process.env.INDRA_NODE_URL;
     const messagingUrl = process.env.INDRA_NATS_URL;
 
-    const randomInterval = Math.round(argv.interval * 0.75 + Math.random() * (argv.interval * 0.5));
-    log.info(`Using random inteval: ${randomInterval}`);
+    const randomInterval = argv.interval;
```
mslipper, line 56: This code is presented here to show how I generated the benchmark numbers. I will revert it prior to merge.

modules/daicard/src/components/sendCard.js
```diff
@@ -123,6 +123,7 @@ export const SendCard = style(
       }
       paymentAction("DONE");
     };
+    window.paymentHandler = paymentHandler;
```
mslipper, line 126: This allows me to instrument the UI if I want to. It will be removed prior to merge.

modules/node/ops/Dockerfile
```diff
@@ -8,4 +8,5 @@ RUN curl https://raw.githubusercontent.com/vishnubob/wait-for-it/ed77b63706ea721
 COPY modules/node/package.json package.json
 RUN npm install --only=production > /dev/null 2>&1
 COPY modules/node /root
+EXPOSE 9229
```
mslipper, line 11: This is required in order to expose the node's debugger port.

modules/node/ops/entry.sh
```diff
@@ -40,21 +40,21 @@ wait_for "nats" "$INDRA_NATS_SERVERS"
...
-else
+# if [[ "$NODE_ENV" == "development" ]]
+# then
+#   echo "Starting indra node in dev-mode"
+#   exec ./node_modules/.bin/nodemon \
+#     --delay 1 \
+#     --exitcrash \
+#     --ignore *.test.ts \
+#     --ignore *.swp \
+#     --legacy-watch \
```
mslipper, line 51: This will be reverted prior to merge.

modules/node/src/appInstance/appInstance.entity.ts
```diff
@@ -124,3 +118,99 @@ export class AppInstance<T extends AppName = any> {
   @UpdateDateColumn()
   updatedAt: Date;
 }
+
+export interface AppInstanceJSON<T extends AppName = any> {
```
mslipper, line 122: I learned that it is possible to instantiate a TypeORM entity without connecting to the database, you just have to manually assign all of the entity's properties. The `*JSON` interfaces defined here and elsewhere represent a JSON-compatible version of the database entity, so that it can be put in Redis properly.

modules/node/src/cfCore/cfCore.store.ts
```diff
@@ -361,6 +373,14 @@ export class CFCoreStore implements IStoreService {
           .execute();
       }
     });
+    await this.cache.mergeCacheValues(`appInstance:identityHash:${freeBalanceAppInstance.identityHash}`, 60000, {
```
mslipper, line 376: This class repeatedly looks up the same data using different criteria. The caching scheme here maps that criteria back to a domain object, and manages purging/merging the cache with new values when the underlying data changes.

modules/node/src/channel/channel.entity.ts
```diff
@@ -88,3 +89,58 @@ export class Channel {
...
+  multisigAddress: string
+  schemaVersion: number
+  addresses: CriticalStateChannelAddresses
+  userIdentifier: string
+  nodeIdentifier: string
+  available: boolean
+  activeCollateralizations: Collateralizations
+  appInstances: AppInstanceJSON[]
+  monotonicNumProposedApps: number
+  // withdrawalCommitments: WithdrawCommitmentJSON[]
```
mslipper, line 103: These fields are not queried via `cfCore`, but should be included for the sake of completeness. I'll add them before this gets merged.

modules/node/src/redis/redis.provider.ts
```diff
@@ -21,6 +21,8 @@ export const redisClientFactory: FactoryProvider = {
   },
 };
 
+const RETRY_DELAY = 50;
```
mslipper, line 24: We can't set `RETRY_DELAY` too low. Setting this too low will result in the backend being bogged down in lock checks.

ops/test/bot.sh
```diff
@@ -51,25 +51,26 @@ else echo "Running in non-interactive mode"
 fi
 
 echo "Starting bot registry container"
-docker run \
-  --detach \
-  --entrypoint="bash" \
-  $interactive \
+  docker run \
+    --detach \
+    -p "$((n + 8330)):9229" \
```
mslipper, line 56: Exposes the registry's debugger at 8331.

ops/test/bot.sh
```diff
@@ -88,6 +89,7 @@ do
   echo "Starting agent container $n"
   docker run \
     --detach \
+    -p "$((n + 9330)):9229" \
```
mslipper, line 92: Exposes each agent's debugger at 9330 + n.

modules/node/src/appInstance/appInstance.entity.ts
rhlsthrm: We have a standard type exported from the `types` module for `AppInstanceJSON` and `StateChannelJSON`. Can we try to reuse those types? They are standard across node and client.

modules/node/src/appRegistry/appRegistry.repository.ts
```diff
@@ -5,10 +5,22 @@ import { AppRegistry } from "./appRegistry.entity";
 @EntityRepository(AppRegistry)
 export class AppRegistryRepository extends Repository<AppRegistry> {
   async findByNameAndNetwork(name: string, chainId: number): Promise<AppRegistry> {
-    return this.findOne({ where: { name, chainId } });
+    return this.findOne({
```
rhlsthrm, line 8: Interesting, these AppRegistry entities are essentially config objects. Maybe it would be better if they were loaded into memory on startup? We don't expect these to ever change at runtime.

modules/node/src/cfCore/cfCore.store.ts
```diff
@@ -1,4 +1,4 @@
-import { Injectable } from "@nestjs/common";
+import {Injectable} from '@nestjs/common';
```
rhlsthrm, line 1: Yikes looks like linter wars. We have a prettier config for this, can you use that?

modules/node/src/channel/channel.entity.ts
rhlsthrm: See earlier comment about `StateChannelJSON` interface in the `types` module. We should try to extend that interface and add the node-specific params.

mslipper: I'll see if I can plug this directly into the protocol.

mslipper:
Closing in favor of the branch PR.

Issue #1086, opened 2020-04-30
[client] Consolidate and fix client `resolveCondition` logic


rhlsthrm:
Error seen in Graph's setup:

```
2020-04-30T10:39:19.627Z +6s error IndexerService → Caught error unlocking transfer, waiting 5 seconds and retrying...: Error: Cannot take action on an app with a finalized state.
2020-04-30T10:39:24.629Z [ResolveSignedTransferController] resolveSignedTransfer started: {
  "conditionType": "SignedTransfer",
  "paymentId": "0x78cf66c74a63185c926a7812dd67f4b5563c4a831bf49a576081667d9668b4dd",
  "data": "0x01d8a6d7b65df6d23191c6ade6aeacf2aba53818d99d44bd0420448a0e418429",
  "signature": "0x222825de4b882ade494e18e2060e0b4230f879e7dc2ea57c60f9e6ef4e0fca5f0e2059be6a39b7215a5662de4ed66e87ccd6876686ea29c1b515402619a70d911c"
}
2020-04-30T10:39:26.085Z [CF-RpcRouter] Processed chan_getAppInstance method in 1450 ms
```

We should directly uninstall if the state is already finalized.

ArjunBhuptani:
Changing this to be a larger issue -- we should just go ahead and properly fix all of the resolve logic here since it's highly custom and messy

PR #1199, opened 2020-06-01
[client] Consolidate transfers

rhlsthrm:
## The Problem
#1086 

## Checklist:


modules/types/src/transfers.ts
```diff
@@ -4,18 +4,40 @@ import {
...
-export const GenericConditionalTransferAppName = "GenericConditionalTransferApp";
+export type TransferType = "RequireOnline" | "AllowOffline";
+export const getTransferTypeFromAppName = (
+  name: SupportedApplications,
+): TransferType | undefined => {
+  switch (name) {
+    case SupportedApplications.DepositApp:
+    case SupportedApplications.SimpleTwoPartySwapApp:
+    case SupportedApplications.WithdrawApp: {
+      return undefined;
```
bohendo, line 22: It's not obvious what an undefined return value means here. If it doesn't apply, we could explicitly return a "NotApplicable" `TransferType`.
Additionally, it seems like we use this fn to tag just a few apps where it *is* relevant, therefore we should default to `undefined` for all others.

Defining exceptions out of existence is recommended by A Philosophy of Software design. The highlights I made while reading this book are available here: https://gitlab.com/bohendo/my-clippings/-/blob/master/highlights/a-philosophy-of-software-design.md
> loc 1827 - This chapter focused on exceptions, which are one of the most significant sources of special-case code, and discussed how to reduce the number of places where exceptions must be handled. The best way to do this is by redefining semantics to eliminate error conditions. For exceptions that can’t be defined away, you should look for opportunities to mask them at a low level, so their impact is limited, or aggregate several special-case handlers into a single more generic handler.

modules/client/src/controllers/ConditionalTransferController.ts
```diff
@@ -0,0 +1,155 @@
...
+  PublicResults,
+  GenericConditionalTransferAppState,
+} from "@connext/types";
+import { toBN, stringify } from "@connext/utils";
+import { HashZero, Zero } from "ethers/constants";
+
+import { AbstractController } from "./AbstractController";
+import { soliditySha256 } from "ethers/utils";
+
+export class ConditionalTransferController extends AbstractController {
```
bohendo, line 17: Calling this a "ConditionalTransfer" Controller suggests that it's distinct from other types of transfer controllers. But, now that we've consolidated all transfers into one controller, the "Conditional" prefix isn't useful for distinguishing this type of transfer logic from others. I'd recommend calling this module "CreateTransferController" & it's sibling "ResolveTransferController" instead.

modules/client/src/connext.ts
```diff
@@ -343,42 +321,19 @@ export class ConnextClient implements IConnextClient {
   public resolveCondition = async (
```
bohendo, line 321: While we're refactoring I'd like to clean up some of these names. It's not obvious what `resolveCondition` does just from it's method name (what kind of condition?). Event the `transfer` method suggests an atomic operation where, in reality, there are two distinct steps. IMO two methods `createTransfer` and `resolveTransfer` are much more obvious. What do you think?

(If we do rename these methods, we should prob keep the old ones around too for backwards-compatibility but mark them as depreciated & flag for removal during our next breaking release)

modules/client/src/controllers/ConditionalTransferController.ts
ArjunBhuptani: I agree with this generally but perhaps we should still have an atomic `transfer` method? We'd still need at least one type of transfer to be the "easy"/default transfer that requires minimal overhead.

modules/client/src/connext.ts
rhlsthrm: I held off on the interface breaking changes. I renamed the methods internally. We can do a separate cleanup around the interface later.

PR #1203, opened 2020-06-02
Performance improvements - part 1

mslipper:
See https://github.com/connext/indra/pull/1193/.


mslipper:
I'll fix the integration tests tonight.

PR #1215, opened 2020-06-05
1204 fix state restore

bohendo:
Doesn't 100% solve any issues yet but does fix one major bug in offline deposit tests.
Also adds extra debugging info for the 1204 bug which will be a big help once deployed.


modules/cf-core/src/cfCore.ts
```diff
@@ -338,7 +343,8 @@ export class CFCore {
             break;
           }
           default: {
-            throw new Error(`Unrecognized persist state channel type: ${type}`);
+            const c: never = type;
```
bohendo, line 346: Tightening up the types a little, I dig it

modules/store/src/store.ts
```diff
@@ -206,6 +206,20 @@ export class StoreService implements IStoreService {
     });
   }
 
+  async incrementNumProposedApps(multisigAddress: string): Promise<void> {
+    return this.execute((store) => {
+      const channel = this.getStateChannelFromStore(store, multisigAddress);
+      if (!channel) {
+        throw new Error(`Can't save app instance without channel`);
```
bohendo, line 213: This error message doesn't quite fit

modules/cf-core/src/testing/scenarios/sync.spec.ts
```diff
@@ -222,6 +224,170 @@ describe("Sync", () => {
...
+      expect(isHexString(multisigAddress)).toBe(true);
+      expectedChannel = await storeServiceA.getStateChannel(multisigAddress);
+      const unsynced = await storeServiceB.getStateChannel(multisigAddress);
+      expect(expectedChannel).toBeDefined();
+      expect(expectedChannel!.proposedAppInstances.length).toBe(0);
+      expect(unsynced?.proposedAppInstances.length).toBe(0);
+      messagingServiceA.clearLimits();
+    });
+
+    test.only("sync protocol responder is missing a proposal held by the protocol initiator", async () => {
```
bohendo: .only?! :scream:

modules/cf-core/src/protocol/sync.ts
```diff
@@ -102,14 +102,26 @@ export const SYNC_PROTOCOL: ProtocolExecutionFlow = {
...
+        const proposalSync = await syncNumProposedApps(postSyncStateChannel, responderChannel);
+        if (proposalSync) {
+          yield [
+            PERSIST_STATE_CHANNEL,
+            PersistStateChannelType.SyncNumProposedApps,
+            proposalSync.updatedChannel,
+          ];
+          postSyncStateChannel = StateChannel.fromJson(proposalSync!.updatedChannel.toJson());
+        }
+        logTime(log, substart, `[${processID}] Synced proposals with initiator`);
```
bohendo, line 115: Should this be "Synced proposals with responder"?

Issue #1217, opened 2020-06-07
[node] Support non-18 decimal tokens


rhlsthrm:
This should be mostly node-side work. We can just use literal units for all inputs, but the node needs to collateralize the tokens with the correct units.

PR #1218, opened 2020-06-07
1217 support non 18 decimal units

rhlsthrm:
## The Problem
#1217 

## The Solution

Update node to convert rebalancing targets to the correct units.

## Checklist:


modules/node/src/test/utils/config.ts
```diff
@@ -15,18 +15,28 @@ export const defaultSigner = new ChannelSigner(
...
+  extraSupportedTokens: string[];
+};
+
 export class MockConfigService extends ConfigService {
-  private nodeSigner;
-  constructor(signer: ChannelSigner = defaultSigner) {
+  private nodeSigner: ChannelSigner;
+  private supportedTokens: string[];
+
+  constructor(overrides: Partial<ConfigOverrides> = {}) {
```
bohendo, line 28: This is a better pattern for customizing the MockConfigService :+1:

PR #1247, opened 2020-06-17
Bot investigation

LayneHaber:
## The Problem
Investigating errors reported via bots. 

## The Solution
Not conclusive, some changes made:
- protocols log with appId/multisigAddress not process id
- responders protocols will fail if it took >5s to execute (ie longer than counterparty would wait for reply)
- set state commitment signatures can be added in any order and they will be automatically sorted (as in conditional commitments). Note that this *does* add extra time, but we try to run this against the graph setup before taking out unnecessary recover calls.

## Checklist:


LayneHaber:
Current status: still some problems with node and collateral in the bot tests. Was able to successfully send 100 payments with 10 bots at 1000ms interval

PR #1300, opened 2020-07-08
Hotfix transfer statuses

LayneHaber:
## The Problem
Transfer statuses have racy bugs

## The Solution
Consolidate logic to determine Transfer status from sender and receiver apps. Also consolidate typedefs

## Checklist:

### DIFF
--- a/modules/node/src/utils.ts
+++ b/modules/node/src/utils.ts
@@ -30,6 +30,10 @@ export function appStatusesToTransferStatus<T extends AppName>(
       // FIXME: How will you be able to determine if this was a
       // collaborative payment cancellation/failure instead of a
       // success?
+      // cant check latest state preimage on app because there is no
+      // guarantee the app state in our db corresponds to the latest
+      // state of the app (updating state to include preimage does not
+      // happen under lock)
       return TransferStatuses.COMPLETED;
     }
     case AppType.INSTANCE: {
--- a/modules/node/src/utils.ts
+++ b/modules/node/src/utils.ts
@@ -1,0 +1,90 @@
+import {
+  TransferWithExpiryStatus,
+  AppName,
+  TransferStatus,
+  TransferStatuses,
+  TransferWithExpiryStatuses,
+  HashLockTransferAppState,
+} from "@connext/types";
+import { AppInstance, AppType } from "./appInstance/appInstance.entity";
+import { bigNumberifyJson } from "@connext/utils";
+
+export function appStatusesToTransferStatus<T extends AppName>(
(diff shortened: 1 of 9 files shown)


modules/node/src/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+    case AppType.PROPOSAL: {
+      return TransferStatuses.PENDING;
+    }
+    case AppType.UNINSTALLED: {
+      if (!receiverApp) {
+        // Sender app was uninstalled and receiver app rejected
+        return TransferStatuses.FAILED;
+      }
+      if (receiverApp.type !== AppType.UNINSTALLED) {
+        // receiver app exists, while sender app has been uninstalled
```
rhlsthrm, line 26: This case is actually impossible to happen, we don't allow the sender to uninstall before the receiver. Maybe we should hard error here?

modules/node/src/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+      // success?
+      // cant check latest state preimage on app because there is no
+      // guarantee the app state in our db corresponds to the latest
+      // state of the app (updating state to include preimage does not
+      // happen under lock)
+      return TransferStatuses.COMPLETED;
+    }
+    case AppType.INSTANCE: {
+      if (!receiverApp) {
+        // FIXME: if a receiver app is rejected, we won't be able to retrieve
```
ArjunBhuptani, line 41: This one is ok --

If an install fails on the receiver side for some reason (e.g. sender finished installing and returned sigs but receiver did not get them), then the sender AppType would be INSTANCE and the receiver AppType would EITHER be:
1. PROPOSAL - in case the receiver hasn't yet realized that the installation failed (alternatively, the installation hasn't yet completed on receiver side)
2. undefined - in the case that the receiver realized that the installation failed and then rejected the proposal

modules/node/src/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+        return TransferStatuses.FAILED;
+      }
+      // sender and receiver have uninstalled their apps
+      // FIXME: How will you be able to determine if this was a
+      // collaborative payment cancellation/failure instead of a
+      // success?
+      // cant check latest state preimage on app because there is no
+      // guarantee the app state in our db corresponds to the latest
+      // state of the app (updating state to include preimage does not
+      // happen under lock)
```
ArjunBhuptani, line 36: You're right - there's no way to really have the status be conditional upon latest state unless that state is updated under the lock. Maybe we *do* have to save the latest state after all...

This shouldn't block merging this PR for now, though.

Issue #1288, opened 2020-07-06
[all] Add Address/ABI context to commitments in store


ArjunBhuptani:
When going through breaking changes, we need to ensure that users _always_ have the ability to dispute with older commitments, even if their network context has changed as part of an update.

Commitments should be paired with relevant addresses/ABIs *from when the commitment was created* so that they can always be disputed after the fact.

PR #1301, opened 2020-07-08
1288 add context to commitments

rhlsthrm:
## The Problem

#1288 

## The Solution

Added tx data to all commitments at time of creation.

## Checklist:


modules/node/src/cfCore/cfCore.store.spec.ts
```diff
@@ -335,7 +335,7 @@ describe("CFCoreStore", () => {
   });
 
   describe("App Instance", () => {
-    it("should not create an app instance if there is no app proposal", async () => {
+    it.skip("should not create an app instance if there is no app proposal", async () => {
```
bohendo, line 338: If this was on purpose, let's add a comment explaining what needs to happen before we unskip this test.

modules/node/src/conditionalCommitment/conditionalCommitment.repository.ts
```diff
@@ -45,14 +46,10 @@ export class ConditionalTransactionCommitmentRepository extends Repository<
   ): Promise<ConditionalTransactionCommitment[]> {
     return this.createQueryBuilder("conditional")
       .leftJoinAndSelect("conditional.app", "app")
-      .where(
-        "app.type <> :rejected", { rejected: AppType.REJECTED },
-      )
+      .where("app.type <> :rejected", { rejected: AppType.REJECTED })
```
bohendo, line 49: Has this been consolidated with Layne's change to remove the rejected app type? I thought rejected apps were deleted & therefore aren't around to have their AppType set to rejected.

modules/contracts/src.ts/commitments/set-state-commitment.ts
```diff
@@ -111,13 +121,15 @@ export class SetStateCommitment implements EthereumCommitment {
...
       bnJson.signatures[0],
       bnJson.signatures[1],
     );
   }
 
-  public async getSignedAppChallengeUpdate(): Promise<SignedAppChallengeUpdate> {
-    await this.assertSignatures();
+  public getSignedAppChallengeUpdate(): SignedAppChallengeUpdate {
+    // TODO: figure out where to do this rather than in this function
+    // await this.assertSignatures();
```
bohendo, line 132: Looks like we're calling the assertion in getSignedTransaction(), so is this TODO still relevant?

modules/node/src/conditionalCommitment/conditionalCommitment.entity.ts
```diff
@@ -2,31 +2,35 @@ import { OneToOne, JoinColumn, Entity, Column, PrimaryColumn } from "typeorm";
 
 import { AppInstance } from "../appInstance/appInstance.entity";
 import { IsEthAddress, IsKeccak256Hash } from "../validate";
+import { HexString } from "@connext/types";
```
bohendo, line 5: Imports don't need to be sorted but let's at least keep external imports separate from local ones

modules/node/src/conditionalCommitment/conditionalCommitment.entity.ts
```diff
@@ -2,31 +2,35 @@ import { OneToOne, JoinColumn, Entity, Column, PrimaryColumn } from "typeorm";
...
 
 @Entity()
 export class ConditionalTransactionCommitment {
   @PrimaryColumn()
   @IsKeccak256Hash()
-  appIdentityHash!: string;
+  appIdentityHash!: HexString;
```
bohendo, line 11: More descriptive string types: nice :sunglasses:

modules/node/src/conditionalCommitment/conditionalCommitment.repository.ts
ArjunBhuptani: That change isn't in staging yet -- lets merge this first since there are a few extra considerations for fixing statuses

ArjunBhuptani:
There's still one remaining issue in this PR:

The stored proc call for `createAppInstance` isn't properly erroring when no app proposal exists. Instead, it just fails silently and returns `undefined`. 

@rhlsthrm and I also found that this was being obscured by the cache set calls in the same function -- which is why calling `getAppInstance` afterwards returned the correct data. It's an unsafe practice to directly write to cache like this without validating the output of the procedure, since it would then be possible to continue on with the protocol without saving any of the commitments.

rhlsthrm:
Fixed the issue @ArjunBhuptani mentioned.

PR #1310, opened 2020-07-11
39 cooperative cancel

LayneHaber:
## The Problem
Fixes [this](https://github.com/connext/rest-api-client/issues/39) in the client/node

## The Solution
Call `resolveCondition` with a payment id and no secret to cancel a payment.

## Checklist:


modules/client/src/controllers/ResolveTransferController.ts
```diff
@@ -210,4 +225,81 @@ export class ResolveTransferController extends AbstractController {
...
+    if (
+      state.coinTransfers[1].to === this.connext.signerAddress &&
+      toBN(state.expiry || 0).gt(block)
+    ) {
+      action = { preImage: HashZero };
+    }
+    this.log.info(`[${paymentId}] Uninstalling transfer app with action ${stringify(action)}`);
+    await this.connext.uninstallApp(paymentApp.identityHash, action);
+    this.log.info(`[${paymentId}] Uninstalled transfer app ${paymentApp.identityHash}`);
+    return ret;
```
ArjunBhuptani, line 303: This file is unnecessarily complicated. You're already doing `if(state.expiry)...`, so you might as well just do this:

```
if(state.expiry && state.expiry > currentBlockNumber) {
   await uninstallApp(appIdentityHash) // uninstall without taking action
} else {
   await uninstallApp(action, appIdentityHash) // uninstall with action - this will cover *all* other cases!!
```

PR #1329, opened 2020-07-22
1153 onchain tx service

### LINKED ISSUE #1153: [cf-core] Transaction Service
**Describe the bug**
The client and node both have special methods for sending transactions, that can be configured for gas, retries with errors, storing info, etc.

Currently, the `chan_deployStateDepositHolder` method sends the transaction itself within the cf-core. Ideally, this method would use the same logic that the node and client use to send all other transactions to chain.

Note: sometimes this will pop up in CD, and waste a lot of time.

**Solution**
`cf-core` should take in a transaction service in the same way it takes in a lock service, and use that to send all transactions.

LayneHaber:
## The Problem
Fixes #1153 

## Checklist:


LayneHaber:
Must merge #1360 first

PR #1341, opened 2020-07-27
E2E bot test

rhlsthrm:
## Checklist:

### DIFF
--- a/modules/bot/src/e2e.ts
+++ b/modules/bot/src/e2e.ts
@@ -1,25 +1,15 @@
-import { artifacts } from "@connext/contracts";
-import {
-  abrv,
-  toBN,
-  ChannelSigner,
-  ColorfulLogger,
-  abbreviate,
-  getRandomPrivateKey,
-} from "@connext/utils";
-import { constants, Contract, providers, utils, Wallet, BigNumber } from "ethers";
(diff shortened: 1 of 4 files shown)


modules/bot/src/e2e.ts
```diff
@@ -0,0 +1,171 @@
...
+        describe: "Asset id for payments",
+        type: "string",
+        default: AddressZero,
+      })
+      .option("chain-id", {
+        describe: "Chain ID to test",
+        type: "number",
+        default: 1337,
+      })
+      .option("funder-mnemonic", {
```
ArjunBhuptani, line 34: We need to make sure to correctly pass in funder-mnemonic because it will have to be the node's mnemonic, not sugar daddy. Perhaps we should even just use whatever's in the env by default?

ops/test/e2e.sh
```diff
@@ -0,0 +1,45 @@
...
+else echo "Running in non-interactive mode"
+fi
+
+exec docker run \
+  $interactive \
+  --entrypoint="bash" \
+  --env="INDRA_CHAIN_URL=$INDRA_CHAIN_URL" \
+  --env="INDRA_NODE_URL=$INDRA_NODE_URL" \
+  --env="LOG_LEVEL=$LOG_LEVEL" \
+  --env="MNEMONIC=$MNEMONIC" \
```
ArjunBhuptani, line 28: Ah ok nvm

PR #1346, opened 2020-07-28
1312 separate ui

### LINKED ISSUE #1312: [ops] Separate UI from node
Current State: Right now, our core Indra Node stack includes the daicard UI. The daicard.io proxy is a transparent proxy for the node at eg indra.connext.network.

Problems:
 - ETC & others dgaf about daicard UI but not easy to turn off in prod
 - UI is served from root path of Indra proxy, this requires an otherwise unnecessary `/api` prefix to access the node.
 - daicard ops are tightly bundled w rest of indra, makes it not as useful as a standalone example integration.

Solution: Separate daicard/dashboard ops from the rest of the node ops. Move anything relevant to the daicard from `ops/start-{dev,prod}` into `ops/start-daicard`. The UI's js/html/css files will live on the server at daicard.io instead of the one at indra.connext.network.

Benefits:
 - Nobody has to worry about the daicard unless they explicitly choose to
 - 1 fewer service in the core node stack but we could easily start up the daicard stack too during `make start` if we want to.
 - The need for an `/api` prefix goes away entirely. Url management & proxy config in general all become simpler.
 - daicard is more useful as a standalone example & could trivially be exported into a separate repo if we ever wanted to.

bohendo:
## The Problem

closes #1312

## Checklist:


modules/daicard/src/App.js
```diff
@@ -35,8 +35,8 @@ const { formatEther } = utils;
 
 const urls = {
   ethProviderUrl:
-    process.env.REACT_APP_ETH_URL_OVERRIDE || `${window.location.origin}/api/ethprovider`,
-  nodeUrl: process.env.REACT_APP_NODE_URL_OVERRIDE || `${window.location.origin}`,
+    process.env.REACT_APP_ETH_URL_OVERRIDE || `${window.location.origin}/ethprovider`,
```
rhlsthrm, line 38: We'll need to fix this for multichain.

modules/contracts/address-book.json
```diff
@@ -1179,135 +1179,7 @@
       "txHash": "0x1e787801c045702c8b8d6bcfea5ec0c3118dd088fb17e21e6248a730dbb6a572"
     },
     "Token": {
```
rhlsthrm, line 1181: We should probably remove the 1337 and 1338 entries from the address book completely since they are in .chaindata now.

modules/daicard/src/App.js
bohendo: Well that'll be a larger overhaul to add multichain support to the daicard so is pretty far outside the scope of this PR..
However, Indra's proxy *should* expose providers for each of it's chains.. I'll look into that

Issue #1359, opened 2020-08-03
[node, client] Return wrapped tx object for deposit/withdraw/collateral


ArjunBhuptani:
Currently, to properly wait on a collateralization request we have to:
```
const tx = await client.requestCollateral(AddressZero) as any;
await ethProvider.waitForTransaction(tx.hash);
await client.waitFor(EventNames.UNINSTALL_EVENT, 10_000);
```
This is kind of ugly & requires arcane knowledge of event names. This waiting process should be simpler but `requestCollateral` still needs to return asap

bohendo:
Returning something like `{ tx: TransactionResponse; finished: Promise<void> }` might be good.

Then the end user can wait on the tx & protocol separately. We need to be careful to document this carefully though so these two things we can wait on don't get mixed up.

PR #1365, opened 2020-08-05
1359 wrapped tx object

rhlsthrm:
## The Problem

Deposit method waits for full tx. #1359 

## The Solution

Deposit return value changed to include awaitable promise that resolves upon completion.

## Checklist:

### DIFF
--- a/modules/client/src/controllers/DepositController.ts
+++ b/modules/client/src/controllers/DepositController.ts
@@ -46,7 +46,6 @@ export class DepositController extends AbstractController {
     this.throwIfAny(notLessThanOrEqualTo(amount, startingBalance), notGreaterThan(amount, Zero));
     const { appIdentityHash } = await this.requestDepositRights({ assetId });
 
-    let ret: PublicResults.RescindDepositRights;
     const txHash = await this.connext.channelProvider.walletDeposit({
       amount: amount.toString(),
       assetId: tokenAddress,
@@ -70,7 +69,7 @@ export class DepositController extends AbstractController {
             amount: amount,
             assetId: tokenAddress,
           } as EventPayloads.DepositConfirmed);
-          this.log.info(`deposit complete for assetId ${assetId}: ${JSON.stringify(ret)}`);
+          this.log.info(`deposit complete for assetId ${assetId}: ${JSON.stringify(res)}`);
           resolve(res);
         } catch (e) {
           this.connext.emit(EventNames.DEPOSIT_FAILED_EVENT, {
@@ -78,7 +77,7 @@ export class DepositController extends AbstractController {
             assetId: tokenAddress,
             error: e.stack || e.message,
           } as EventPayloads.DepositFailed);
-          ret = await this.rescindDepositRights({ appIdentityHash, assetId });
+          await this.rescindDepositRights({ appIdentityHash, assetId });
           reject(new Error(e.stack || e.message));
         }
       },
--- a/modules/client/src/controllers/DepositController.ts
+++ b/modules/client/src/controllers/DepositController.ts
@@ -47,43 +47,46 @@ export class DepositController extends AbstractController {
     const { appIdentityHash } = await this.requestDepositRights({ assetId });
 
     let ret: PublicResults.RescindDepositRights;
-    let transactionHash: string;
-    try {
-      this.log.debug(`Starting deposit`);
-      transactionHash = await this.connext.channelProvider.walletDeposit({
-        amount: amount.toString(),
-        assetId: tokenAddress,
-      });
-      this.connext.emit(EventNames.DEPOSIT_STARTED_EVENT, {
-        amount: amount,
-        assetId: tokenAddress,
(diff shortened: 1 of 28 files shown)


modules/test-runner/src/deposit/deposit.test.ts
```diff
@@ -68,7 +68,11 @@ describe("Deposits", () => {
       client: ONE,
       assetId: AddressZero,
     };
-    await client.deposit({ amount: expected.client, assetId: expected.assetId });
+    const response = await client.deposit({ amount: expected.client, assetId: expected.assetId });
+    if (response.completed) {
+      expect(response.txHash).to.be.ok;
```
bohendo, line 73: Would it be feasible for client.deposit to return the entire transaction instead of just the hash? bc that'd frequently save an `await ethProvider.getTransaction(txHash)` step

modules/test-runner/src/deposit/deposit.test.ts
```diff
@@ -219,9 +239,20 @@ describe("Deposits", () => {
       assetId: tokenAddress,
       node: Zero,
     };
-    await client.deposit({ amount: TWO, assetId: ethExpected.assetId });
+    let response = await client.deposit({ amount: TWO, assetId: ethExpected.assetId });
+    if (response.completed) {
```
bohendo, line 243: In what cases would we *not* have a `response.completed` here?

modules/types/src/public.ts
```diff
@@ -12,12 +12,17 @@ type DepositParameters = {
   assetId?: Address; // if not provided, will default to 0x0 (Eth)
 };
 
-type DepositResponse = {
+type FreeBalanceResponse = {
```
bohendo, line 15: Much better name :pray:

modules/types/src/public.ts
```diff
@@ -12,12 +12,17 @@ type DepositParameters = {
...
   freeBalance: {
     [s: string]: BigNumber;
   };
 };
 
+type DepositResponse = {
+  txHash: string;
+  completed: () => Promise<FreeBalanceResponse>;
```
bohendo, line 23: Having the deposit promise resolve to the new free balance is cool.
It makes me wonder if we ought to change the name of that promise to indicate that it contains useful info, maybe `balanceUpdated` or `freeBalanceUpdateComplete`
idk I'm just being a stickler about the name of this promise lol it's not that important.

modules/test-runner/src/deposit/deposit.test.ts
rhlsthrm: Yes this is definitely doable.

modules/test-runner/src/deposit/deposit.test.ts
rhlsthrm: Backwards compatibility. In old clients, the deposit method would wait for the full deposit to complete.

modules/types/src/public.ts
rhlsthrm: Hmm, I'm inclined to agree but my issue is that it's a little misleading. It's not _just_ the balance being updated, the channel isn't usable until this process is complete, that's why `completed` makes more sense to me.

modules/client/src/controllers/DepositController.ts
```diff
@@ -47,43 +47,46 @@ export class DepositController extends AbstractController {
...
+          });
+          this.log.info(`Sent deposit transaction to chain: ${txHash}`);
+          await transaction.wait();
+          const res = await this.rescindDepositRights({ appIdentityHash, assetId });
+          this.connext.emit(EventNames.DEPOSIT_CONFIRMED_EVENT, {
+            hash: txHash,
+            amount: amount,
+            assetId: tokenAddress,
+          } as EventPayloads.DepositConfirmed);
+          this.log.info(`deposit complete for assetId ${assetId}: ${JSON.stringify(ret)}`);
```
LayneHaber, line 73: It doesnt look like ret is assigned to the right variable here

modules/node/src/channel/channel.service.ts
```diff
@@ -188,7 +194,13 @@ export class ChannelService {
...
       `Rebalance finished for ${channel.userIdentifier} on chain ${channel.chainId}, assetId: ${assetId}`,
     );
-    return response;
+    return (
+      depositRes && {
+        transaction: depositRes.tx,
+        depositAppIdentityHash: depositRes.appIdentityHash,
+        rebalanceType,
+      }
+    );
```
LayneHaber, line 203: will this properly return `undefined` or will it return `false`?

rhlsthrm: If `depositRes` is undefined when it's set above (which it should be), it will hit the first half of the statement, and since undefined is falsy, it'll short circuit and return undefined there.

modules/client/src/controllers/DepositController.ts
rhlsthrm: You're right, fixed.

PR #1406, opened 2020-08-27
✨ Add new collateralization config and admin endpoints

### LINKED ISSUE #1405: Pre-Collateralization
**Is your feature request related to a problem? Please describe.**
Mainnet usage has shown, that "on-the-fly" collateralization is not a great long-term approach since it relies on getting a transaction into the next block. Not only is accurate gas estimation for the next block difficult, but also very expensive. In practice we are experiencing confirmation times of several minutes and timeouts as a result. Further, we'd need to be able to set the desired amount for the collateralization since the current default of e.g. 20 USDT triggers a live collateralization (and hence a swap timeout) for all trades of greater 20 USDT

**Describe the solution you'd like**
We think pre-collateralizing a channel with an estimated amount is the better way. Since the indra node doesn't know anything about trades/estimated amounts as of now, we think it makes sense for the client to provide this information and pass it in the collateralization request.

- add amount field collateralization call `(  export type PostRequestCollateralRequestParams = MultiClientRequestParams & { assetId: string };`), this allows xud to request collateral for the required amount via the client before/when an order is entering the order book based on the order amount
- to avoid draining node funds, we would need to be able to set a fixed upper limit per currency per client for the collateralization in the node
- the client would need to get notified with the collateralization tx hash from the node the same way currently deposits do, collateralization status or simply the remote balance needs to be exposed via the rest-api

rhlsthrm:
## The Problem

Fixed #1405 

## Checklist:


docs/src/reference/node.md
```diff
@@ -0,0 +1,29 @@
+# Node
```
LayneHaber, line 1: niceeee

modules/node/src/config/config.service.ts
```diff
@@ -293,32 +297,55 @@ export class ConfigService implements OnModuleInit {
...
+        ...defaultProfileEth,
       };
     }
+    let defaultProfileToken = {
+      collateralizeThreshold: parseEther(`5`),
+      target: parseEther(`20`),
+      reclaimThreshold: parseEther(`100`),
+    };
+    try {
+      defaultProfileToken = JSON.parse(this.get("INDRA_DEFAULT_REBALANCE_PROFILE_TOKEN"));
```
LayneHaber, line 330: what if you support the use of multiple tokens? could we make the `INDRA_DEFAULT_REBALANCE_PROFILE` an array keyed on assetId?

rhlsthrm: I dunno, I thought the same and figured it would get complicated, so this is basically going to set it for all tokens. We can revisit if this becomes a problem. Also at that point, node operators might have to start setting rebalance profiles per channel.

PR #1418, opened 2020-09-05
1417 separate flows

### LINKED ISSUE #1417: [node] Separate offline and online transfers
Put into separate flows to detect + address edgecases more easily.

LayneHaber:
Fixes #1417 

Requires in depth review before merge.

### DIFF
--- a/modules/node/src/transfer/transfer.service.ts
+++ b/modules/node/src/transfer/transfer.service.ts
@@ -152,8 +152,111 @@ export class TransferService {
     );
   }
 
+  // NOTE: This function is called by `transferAppInstallFlow` and 
+  // should hard error if it cannot install the sender app *ONLY*
+  // so it is properly rejected
+  // Otherwise, it should error gracefully from the function, including
+  // performing any relevant cleanup
+  private async offlineTransferFlow(
+    paymentId: string,
+    senderProposal: AppInstance,
+    proposeParams: MethodParams.ProposeInstall,
+    from: string,
+    transferType: ConditionalTransferTypes,
+  ) {
+    // Get the senders channel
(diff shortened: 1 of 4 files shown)


modules/node/src/transfer/transfer.service.ts
```diff
@@ -152,8 +153,165 @@ export class TransferService {
...
+      //   senderProposal.identityHash,
+      //   senderChannel,
+      //   getCancelAction(transferType),
+      // );
+      return;
+    }
+
+    if (
+      !receiverProposeRes ||
+      !receiverProposeRes.appIdentityHash ||
```
rhlsthrm, line 249:
```suggestion
      !receiverProposeRes?.appIdentityHash ||
```

modules/node/src/transfer/transfer.service.ts
```diff
@@ -152,8 +153,165 @@ export class TransferService {
...
+      `Installing receiver app ${receiverProposeRes.appIdentityHash} in channel ${receiverChannel.multisigAddress}`,
+    );
+    let receiverAppInstalled: undefined | AppInstance;
+    try {
+      await this.cfCoreService.installApp(receiverProposeRes.appIdentityHash, receiverChannel);
+      this.log.info(
+        `Receiver app ${receiverProposeRes.appIdentityHash} in channel ${receiverChannel.multisigAddress} installed`,
+      );
+      // Add the receiver app to the transfer
+      receiverAppInstalled = await this.appInstanceRepository.findByIdentityHashOrThrow(
```
rhlsthrm, line 274: Didn't we want to do this in the middleware so it's under the lock?

rhlsthrm (changes requested):
Want to make sure the adding to transfer database model is safe to do outside middleware.

rhlsthrm (approved):

